Issue #7, opened 2019-05-31
Cannot build from source


GabMus:
Here's what I tried to do (note, I don't want to install system wide, I want to build for testing only)

```
$ meson buiThe Meson build system
Version: 0.50.1
Source dir: /home/gabmus/git/foliate
Build dir: /home/gabmus/git/foliate/build
Build type: native build
Project name: com.github.johnfactotum.Foliate
Project version: 1.0.1
Build machine cpu family: x86_64
Build machine cpu: x86_64
Program desktop-file-validate found: YES (/usr/bin/desktop-file-validate)
Program appstream-util found: YES (/usr/bin/appstream-util)
Program glib-compile-schemas found: YES (/usr/bin/glib-compile-schemas)
Found pkg-config: /usr/bin/pkg-config (1.6.1)
Program gjs found: YES (/usr/bin/gjs)
WARNING: Project targetting '>= 0.40.0' but tried to use feature introduced in '0.50.0': install arg in configure_file
Configuring com.github.johnfactotum.Foliate using configuration
Program build-aux/meson/postinstall.py found: YES (/home/gabmus/git/foliate/build-aux/meson/postinstall.py)
Build targets in project: 7
WARNING: Project specifies a minimum meson_version '>= 0.40.0' but uses features which were added in newer versions:
 * 0.50.0: {'install arg in configure_file'}
Found ninja-1.9.0 at /usr/bin/ninjald --prefix=/home/gabmus/git/foliate/fakeusr
$ cd build
$ ninja
[3/3] Generating com.github.johnfactotum.Foliate.appdata.xml_data_merge with a custom command.
$ ninja install
[0/1] Installing files.
Installing subdir /home/gabmus/git/foliate/src/assets to /home/gabmus/git/foliate/fakeusr/share/com.github.johnfactotum.Foliate/assets
Installing /home/gabmus/git/foliate/src/assets/cheerio.js to /home/gabmus/git/foliate/fakeusr/share/com.github.johnfactotum.Foliate/assets
Installing /home/gabmus/git/foliate/src/assets/jszip.min.js to /home/gabmus/git/foliate/fakeusr/share/com.github.johnfactotum.Foliate/assets
Installing /home/gabmus/git/foliate/src/assets/epub.js to /home/gabmus/git/foliate/fakeusr/share/com.github.johnfactotum.Foliate/assets
[...]

PR #8, opened 2019-05-31
Added vim shortcuts

GabMus:
NOTE: I couldn't test this since I wasn't able to build foliate from source. See #7 for more details on this issue.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -1373,6 +1373,8 @@ class BookViewerWindow {
         this.headerBar.pack_end(button)
         this.accelGroup.connect(Gdk.KEY_F, Gdk.ModifierType.CONTROL_MASK, 0, () =>
             button.active = !button.active)
+        this.accelGroup.connect(Gdk.KEY_slash, 0, 0, () =>
+            button.active = !button.active)
     }
     buildView(...args) {
         const button = new Gtk.MenuButton({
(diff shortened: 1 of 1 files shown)


johnfactotum:
I'm not entirely sure if Vim shortcuts are appropriate for GTK applications. Vim shortcuts are more suitable for the CLI or apps that have complex editing/navigating capacities. It'd be better to stick to standard GTK or desktop style shortcuts, in my opinion.

GabMus:
I'm pretty sure there are some gtk apps that implement vim shortcuts.
Besides, my PR doesn't remove existing shortcuts :)

GabMus:
#7 closed, I was able to test the changes I made, and everything seems to be in order

GabMus:
I'm not sure if it's appropriate to write it here, but I wanted to thank you for the amazing work that you're doing on Foliate, keep it up!

johnfactotum:
I've made the shortcuts more consistent with Evince with fc30c1cc4744158fbc7f14b74106119aa6b822cd, this includes n and p for turning pages, and j and k for scrolling in scrolled mode. And also fixed the problem with space conflicting with scroll down in scrolled mode.

So I'll be closing this PR. Thank you for your contribution.

GabMus:
I see that in 305ac2d64e566cfee3c1f52008f42bbeb894a890 you added all the vim shortcuts I implemented, namely `H` for left, `L` for right and `/` for search. They do not show up in the shortcuts tho. Just out of curiosity, what made you change your mind about them?

johnfactotum:
I mentioned it in my comment above:

>  I've made the shortcuts more consistent with Evince...

Them not showing up in the shortcuts window is also in line with Evince. Also in general the shortcuts window has limited space, and it's best not to flood it with too many shortcuts. As mentioned in the [GTK3 Reference Manual](https://developer.gnome.org/gtk3/stable/GtkShortcutsShortcut.html):

> Multiple accelerators can be specified by separating them with a space, but keep in mind that the available width is limited.

GabMus:
I personally think they should be there since they're implemented, but 
that's more of a GTK problem than Foliate's. At the end of the day, 
choice is yours. Thanks for implementing them anyway :)

johnfactotum:
Thanks for the suggestion. I think I'll leave them out for now..

PR #12, opened 2019-05-31
Add AUR repo to README.md

schw0reismus:
I've made a couple of AUR repos for the program. By the way, I added only ```i686``` and ```x86_64``` architectures there because I don't know whether it's gonna work on any kind of ARM procesors.


johnfactotum:
Oh wow, thanks a lot for this! I'll add them to the website later as well.

I wonder if it'd be better to simply link to the package on AUR. Arch Linux users installing from the AUR should know what they're doing. No need to list the detailed commands.

johnfactotum:
I've added them myself with a more concise format.

Thanks again for making the packages!

PR #35, opened 2019-06-03
Update readme to include Void linux instructions

not-chicken:
While building package, gjs wasn't listed as runtime dependency so users have to install it alongside for the time being.

I will fix this in next revision of the package, and will open a PR to remove gjs from instructions.


not-chicken:
Sleepy me did something wrong, will open another pr.

PR #36, opened 2019-06-03
Add Void Linux instructions to Readme

not-chicken:
While building package, gjs wasn't listed as runtime dependency so users have to install it alongside for the time being.

I will fix this in next revision of the package, and will open a PR to remove gjs from instructions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,9 @@ Website: https://johnfactotum.github.io/foliate/
 
 For Arch Linux users, two packages are available on the AUR: [`foliate`](https://aur.archlinux.org/packages/foliate/) (stable version), [`foliate-git`](https://aur.archlinux.org/packages/foliate-git/) (Git version)
 
-For Void Linux users, foliate is available in Void's official repository. Install using xbps: `xbps-install -S foliate gjs`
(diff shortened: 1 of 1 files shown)


johnfactotum:
Thanks a lot for this!

Will it take long to fix the package? If not, I think I'll merge this after it's fixed.

not-chicken:
Thanks to you for creating foliate :+1: , I just packaged it( that too with this little issue).

I was thinking of fixing it along with next update, or a week later. It just got added to repos so I don't want to ping void maintainers so soon.

My vote would be to merge this for the time being. Decision's up to you.

johnfactotum:
Fair enough. I'd appreciate it if you could add an explanation to this in the readme, though, if this is to be merged now. I mean just to provide a slight reminder that `gjs` is a dependency and not something to be explicitly installed.

In fact, does Void Linux's package manager support installing a package as a dependency? If so it might be better to instruct people to do that.

If all this would be too verbose than I think I'd merge after it's fixed.

not-chicken:
I have added a separate note for gjs. Later on we would just have to delete this note.

PR #42, opened 2019-06-04
Fixed translations in font menu

schw0reismus:


johnfactotum:
Did you make these manually or did you somehow get gettext to work?

In any case it's still missing some strings, like the search results ones (not sure if there's anything else).

Issue #40, opened 2019-06-04
The font menu is not fully translated


schw0reismus:
![изображение](https://user-images.githubusercontent.com/51238434/58880946-c5742e80-86e1-11e9-9a00-d2a69829d9e6.png)

johnfactotum:
This is not the only place where gettext fails to extract the strings. I believe this is caused by https://savannah.gnu.org/bugs/?50920. It says the fix is contained in gettext-0.20, but I don't that version at the moment so I haven't tested yet.

shibotto:
Actually it seems even worse with gettext-0.20.1.
Used `ninja com.github.johnfactotum.Foliate-pot` and diffed with 1fe9263f, to keep the output "short":
![immagine](https://user-images.githubusercontent.com/3816409/59103717-ef229500-892f-11e9-94f6-495038841a5c.png)

johnfactotum:
Basically, xgettext seems to go wrong whenever there's a slash inside backticks. My guess is that while they've added support for template strings, the problem with slash in backticks wasn't fixed so it actually made things worse.

I've added a workaround in a37760f745992376ab3d10e80bf9e55de2829ad3, and it works in gettext 0.19.8.1.

johnfactotum:
Closing as the workaround works for now. Should probably report the issue upstream.

PR #54, opened 2019-06-08
FR translation

julroy67:


johnfactotum:
Thanks for this!

I've just added a workaround fixing some strings not being extracted (see #40), can you update the translations?

julroy67:
Not sure if the merge was the right way to do it, but nonetheless here are the new translations. :smile:

PR #57, opened 2019-06-10
Update Italian translation

shibotto:


johnfactotum:
Just added "Quit" and "Margin". Can you add those?

shibotto:
Should be fine :D

PR #62, opened 2019-06-12
Update Russian translation

schw0reismus:


johnfactotum:
Would it be possible to better handle the multiple plural forms? I think it's supported by gettext. You just need to add the plural forms header.

See https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html

schw0reismus:
Thank you very much for your advice! Looks like it works well.

Issue #44, opened 2019-06-04
Please provide an AppImage for Linux


probonopd:
> If a Snap app, PPA, or an AppImage should pop up after I publish this, do let me know so I can add it in.

https://www.omgubuntu.co.uk/2019/05/foliate-ebook-reader-linux

Providing an [AppImage](http://appimage.org/) would have, among others, these advantages:
- Applications packaged as an AppImage can run on many distributions (including Ubuntu, Fedora, openSUSE, CentOS, elementaryOS, Linux Mint, and others)
- One app = one file = super simple for users: just download one AppImage file, [make it executable](http://discourse.appimage.org/t/how-to-make-an-appimage-executable/80), and run
- No unpacking or installation necessary
- No root needed
- No system libraries changed
- Works out of the box, no installation of runtimes needed
- Optional desktop integration with `appimaged`
- Optional binary delta updates, e.g., for continuous builds (only download the binary diff) using AppImageUpdate
- Can optionally GPG2-sign your AppImages (inside the file)
- Works on Live ISOs
- Can use the same AppImages when dual-booting multiple distributions
- Can be listed in the [AppImageHub](https://appimage.github.io/) central directory of available AppImages
- Can double as a self-extracting compressed archive with the `--appimage-extract` parameter
- No repositories needed. Suitable/optimized for air-gapped (offline) machines
- Decentralized

[Here is an overview](https://appimage.github.io/apps) of projects that are already distributing upstream-provided, official AppImages.

If you have questions, AppImage developers are on #AppImage on irc.freenode.net.

johnfactotum:
I'm interested in providing an AppImage, but I'm also at a complete loss as to how to make one.

Foliate uses Meson and follows the [GJS Application Packaging Specification](https://gjs-guide.gitlab.io/guides/gtk/application-packaging.html). With Flatpak, this pretty much _just works_, which is why I was able to get it on Flathub very quickly and easily. Making Snaps, debs, and AppImages is much more involved, particularly because there aren't any templates that I can follow and also of the fact that I don't use Ubuntu personally, and I'm not very familiar with it.

I'd appreciate it if anyone can provide any help with this.

itprojects:
@johnfactotum Since this thread is under "help wanted", the following bash script has been made. It produces binary .deb files which have been tested and work. Minimal effort required!
[package.deb.sh.txt](https://github.com/johnfactotum/foliate/files/3343446/package.deb.sh.txt)

itprojects:
[...]

PR #95, opened 2019-07-01
Added .deb producing script for ninja

itprojects:
Added the script package.deb.bin.sh to be run on demand with ninja like this:

ninja debian-bin

It will make foliate_x.y-z.deb file in build folder.

### DIFF
--- a/build-aux/meson/package.deb.bin.sh
+++ b/build-aux/meson/package.deb.bin.sh
@@ -1,0 +1,78 @@
+#!/bin/bash
+
+# This script produces .deb files from built meson folder
+#
+# cd foliate-master
+# meson build
+# cd build
+# ninja
+# ninja debian-bin
+#
+# Debian/Ubuntu Binary Packaging Stage in this file
+
+# Format foliate version for Debian/Ubuntu, format 1.1.1 to 1.1-1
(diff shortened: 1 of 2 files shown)


johnfactotum:
1. The part after the dash of the version is `debian_revision`, and it "specifies the version of the Debian package based on the upstream version". 1.3.1 should become 1.3.1-1 or similar, not 1.3-1
2. Like I said in #44, redoing everything that's currently done by Meson by hand is unsatisfactory. Maybe use `meson build --prefix=$fnamefull/usr` instead of using `cp` to install things.

itprojects:
1. Fixed the version numbers to just be 1.3.1 format, and not 1.3-1.
2. Yes, it's not satisfactory to copy by hand. On the other hand if the alternative is no package, that's a pool of millions of people that can enjoy your app. Especially for the main linux distributions Ubuntu/Debian/Elementary. Some will no doubt be dedicated maintainers and maybe after the first few .deb's they'll help package it on launchpad or similar. You've got to start somewhere!

johnfactotum:
I've added the specification (e75f79e964c7d637374c3122b01a669bbd376f76). Debian packages can be built using the `dpkg-buildpackage` command.

itprojects:
Thanks, this works.

Issue #98, opened 2019-07-02
Option for selection without default for note


mlouielu:
Hi there,

After testing many e-readers for Linux/epub with lots of frustration, I found foliate, and it works quite good for myself. Deeply thanks to you @johnfactotum.

Currently, foliate's selection is default to adding a note on it. But in my scenario, I sometimes use cursor selection to read through what I'm reading, what I have thought is:

1. Make an option to enable/disable the current function
2. Let right press dragging as selection without adding a new note

I'm familiar with GTK, but not with javascript one, is this possible to do with?

Thanks!

itprojects:
+1 Same here.

johnfactotum:
Now it shows a popover to offer to highlight/lookup/copy instead of doing highlight/lookup on selection. 

The selection is also now cleared when the popover pops down, so it now takes one instead of two clicks to clear the selection.

An additional gain is that it's now possible to lookup phrases instead of just single words.

Will add the option to restore the previous behavior later.

itprojects:
Works well. The user needs to remember to delete the files in the folder ~/.local/share/com.github.johnfactotum.Foliate/* before upgrading to the new code. Otherwise pop-ups might not show.

johnfactotum:
Deleting files shouldn't be necessary. If so, then it's a bug.

itprojects:
It's more likely that the install in question was broken.

PR #99, opened 2019-07-03
Add option to disable fast annotation (impl #98)

mlouielu:
This PR impl #98 

Idea is, when selection appear in `contents.document.onmouseup`, save the `cfiRange` as `latestSelectionRange` and not add into annotation at this round. When user trigger `onmouseup` event again, check if current selection range is same as `latestSelectionRange`, if so, then add it into annotations.

Second commit add the option in menu to enable/disable this function

### DIFF
--- a/data/com.github.johnfactotum.Foliate.gschema.xml
+++ b/data/com.github.johnfactotum.Foliate.gschema.xml
@@ -66,6 +66,11 @@
             <summary>Lookup enabled</summary>
             <description>If true, dictionary lookup is enabled</description>
         </key>
+        <key name="fastannotation-enabled" type="b">
+            <default>true</default>
+            <summary>Fast Annotation enabled</summary>
+            <description>If true, selection as annotation is enabled</description>
+        </key>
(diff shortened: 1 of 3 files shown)


johnfactotum:
Thanks for the PR!

But two things:

1. I'm actually beginning to rework the whole annotation and lookup code. The idea is to make things less coupled and allow adding custom actions (e.g., local dictionaries, translation, web search), and allow the user to configure the default action when something is selected (probably [1] perform an action like lookup or highlight; [2] do nothing but show popover; [3] show popover only on right click) and probably separate configurations for single word and multiple words. It would involve some work in the preference window as well -- that part is currently quite messy.
2. Currently there's a lot of trailing spaces in the code, mostly leftovers from when I was still using an editor that doesn't strip them automatically, and I never got around to fix them. Should probably use a separate commit/PR to remove them to get a cleaner git history.

So thanks again for openng this PR, but I won't be merging it.

mlouielu:
Great! wait for your rework!

PR #130, opened 2019-07-14
Add BibTeX export support

itprojects:
Altered code produces tested bibliographic export of highlights and personal notes, in the Bibtex [.bib] file format.

### DIFF
--- a/src/main.js
+++ b/src/main.js
@@ -366,6 +366,15 @@ const exportToTxt = data => `${_('Annotations for\n%s\nBy %s')
 ${ngettext('%d Annotation', '%d Annotations', data.annotations.length).
 format(data.annotations.length)}
 ${data.annotations.map(({ value, text, color, note }) => `
+--------------------------------------------------------------------------------
+
+${_('Text:')}
+${text}
+${note ? `
+${_('Note:')}
+${note}
+` : ''}`).join('')}
+`
 
 
 const exportToBibTex = data => {
@@ -442,16 +451,6 @@ const exportToBibTex = data => {
 
     return _contents
 }
---------------------------------------------------------------------------------
-
-${_('Text:')}
-${text}
-${note ? `
-${_('Note:')}
-${note}
-` : ''}`).join('')}
-`
-
 
 class Navbar {
     constructor(ttsButton, onSlide, onPrev, onNext, onBack) {
--- a/src/main.js
+++ b/src/main.js
@@ -376,69 +376,68 @@ const exportToBibTex = data => {
         .map(c => {
             switch (c) {
                 case '#':
-                    return '\\#';
+                    return '\\#'
                 case '$':
-                    return '\\$';
+                    return '\\$'
                 case '%':
-                    return '\\%';
+                    return '\\%'
                 case '&':
-                    return '\\&';
+                    return '\\&'
                 case '\\':
-                    return '\\textbackslash{}';
+                    return '\\textbackslash{}'
                 case '^':
-                    return '\\textasciicircum{}';
+                    return '\\textasciicircum{}'
                 case '_':
-                    return '\\_';
+                    return '\\_'
                 case '{':
-                    return '\\{';
+                    return '\\{'
                 case '}':
-                    return '\\}';
+                    return '\\}'
                 case '~':
-                    return '\\textasciitilde{}';
+                    return '\\textasciitilde{}'
                 default:
-                    return c;
+                    return c
             }
     }).join('');
 
     // When missing parameters for author, title, publisher, year
-    var bib_header = '@book{ref'
-        + Math.round(Math.random()*7000).toString() + ':%d' + ',\n';
+    var bib_header = '@book{ref' + Math.round(Math.random()*7000).toString() + ':%d' + ',\n'
 
     if (data.metadata.creator != null && typeof data.metadata.creator !== "undefined") {
-        bib_header += 'author = {%s},\n'.format(badCharReplace(data.metadata.creator));
+        bib_header += 'author = {%s},\n'.format(badCharReplace(data.metadata.creator))
     } else {
-        bib_header += 'author = {unknown},\n';
+        bib_header += 'author = {unknown},\n'
     }
 
     if (data.metadata.title != null && typeof data.metadata.title !== "undefined") {
-        bib_header += 'title = {%s},\n'.format(badCharReplace(data.metadata.title));
+        bib_header += 'title = {%s},\n'.format(badCharReplace(data.metadata.title))
     } else {
-        bib_header += 'title = {unknown},\n';
+        bib_header += 'title = {unknown},\n'
     }
 
     if (data.metadata.publisher != null && typeof data.metadata.publisher !== "undefined") {
-        bib_header += 'publisher = {%s},\n'.format(badCharReplace(data.metadata.publisher));
+        bib_header += 'publisher = {%s},\n'.format(badCharReplace(data.metadata.publisher))
     } else {
-        bib_header += 'publisher = {unknown},\n';
+        bib_header += 'publisher = {unknown},\n'
     }
 
     if (data.metadata.pubdate != null
         && typeof data.metadata.pubdate !== "undefined") {
         bib_header += 'year = %s,\n'
-            .format(data.metadata.pubdate.slice(0, 4));
+            .format(data.metadata.pubdate.slice(0, 4))
     } else {
-        bib_header += 'year = -1,\n';
+        bib_header += 'year = -1,\n'
     }
 
-    bib_header += 'note = {%s}\n},\n\n';
+    bib_header += 'note = {%s}\n},\n\n'
 
-    var annot = '';
+    var annot = ''
     for (var i = 0;i < data.annotations.length;++i) {
-        annot = _("Quote: ") + '"%s"'.format(data.annotations[i].text);
+        annot = _("Quote: ") + '"%s"'.format(data.annotations[i].text)
         if (typeof data.annotations[i].note !== 'undefined') {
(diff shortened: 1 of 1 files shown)


johnfactotum:
Can you follow the existing coding style? 4 spaces for indentation and no semicolons. Also perhaps move this into its own function like the HTML and txt export code.

itprojects:
@johnfactotum Request updated, please test.

johnfactotum:
Have you tested this yourself? You seem to have inserted the function in the middle of the `exportToTxt` function.

src/main.js
```diff
@@ -377,6 +377,81 @@ ${note}
...
+                    return '\\textasciitilde{}'
+                default:
+                    return c
+            }
+    }).join('');
+
+    // When missing parameters for author, title, publisher, year
+    var bib_header = '@book{ref' + Math.round(Math.random()*7000).toString() + ':%d' + ',\n'
+
+    if (data.metadata.creator != null && typeof data.metadata.creator !== "undefined") {
```
johnfactotum: I think a simple `if (data.metadata.creator)` should suffice here.

src/main.js
```diff
@@ -377,6 +377,81 @@ ${note}
...
+        bib_header += 'author = {%s},\n'.format(badCharReplace(data.metadata.creator))
```
johnfactotum: Should use template strings instead of `String.prototype.format`. The latter is only needed when working with gettext.

src/main.js
```diff
@@ -377,6 +377,81 @@ ${note}
 `
 
 
+const exportToBibTex = data => {
+    var _contents
```
johnfactotum: Use `let` instead of `var`. The underscore is unnecessary.

src/main.js
johnfactotum: Should remove the brackets to conform to the coding style. Or better, use the ternary operator instead of `if` statements to make it more concise.

src/main.js
```diff
@@ -377,6 +377,81 @@ ${note}
...
+        bib_header += 'year = %s,\n'
+            .format(data.metadata.pubdate.slice(0, 4))
+    } else {
+        bib_header += 'year = -1,\n'
+    }
+
+    bib_header += 'note = {%s}\n},\n\n'
+
+    var annot = ''
+    for (var i = 0;i < data.annotations.length;++i) {
```
johnfactotum: Use `Array.prototype.map` instead of `for` loops.

src/main.js
```diff
@@ -377,6 +377,81 @@ ${note}
 `
 
 
+const exportToBibTex = data => {
+    var _contents
+
+    // Escape all Tex characters that BibTex requires
+    const badCharReplace = cont => cont.split('')
```
johnfactotum: It's better to use `Array.from` instead of `split`, in my opinion.

src/main.js
```diff
@@ -377,6 +377,81 @@ ${note}
...
+                    return '\\}'
+                case '~':
+                    return '\\textasciitilde{}'
+                default:
+                    return c
+            }
+    }).join('');
+
+    // When missing parameters for author, title, publisher, year
+    var bib_header = '@book{ref' + Math.round(Math.random()*7000).toString() + ':%d' + ',\n'
```
johnfactotum: Not very familiar with BibTex. What is the significance of the number `Math.round(Math.random()*7000)` here? Why random? Why 7000?

johnfactotum:
You changed 7000 to 10000, but what does it mean? Can you add a comment to explain that bit?

It can be made much more concise like:
```javascript
const exportToBibTex = data => {
    // Escape all Tex characters that BibTex requires
    const esc = cont => !cont ? '' : Array.from(cont).map(c =>
        c === '#' ? '\\#'
        : c === '$' ? '\\$'
        : c === '%' ? '\\%'
        : c === '&' ? '\\&'
        : c === '\\' ? '\\textbackslash{}'
        : c === '^' ? '\\textasciicircum{}'
        : c === '_' ? '\\_'
        : c === '{' ? '\\{'
        : c === '}' ? '\\}'
        : c === '~' ? '\\textasciitilde{}'
        : c).join('')

    return data.annotations.map(({ text, note }) =>
`@book{ref${Math.round(Math.random()*10000)}:%d,
    author = {${esc(data.metadata.creator) || 'unknown'}},
    publisher = {${esc(data.metadata.publisher) || 'unknown'}},
    year = {${data.metadata.pubdate ? data.metadata.pubdate.slice(0, 4) : 'unknown'}},
    note = ${_('Quote: ') + text + (note ? _(" Note: ") + note : ''))}
},
`).join('')
}
```

itprojects:
As mentioned in one of the comments above:

> `Math.round(Math.random()*10000)` is necessary to prevent index collisions. Please leave it like it is!

**This is from experience:**
A unique book string variable is a must. The word ref plus a number (3-4 digits) can easily prevent BibTex problems, and content confusion (Was this the right quote for Chapter 1 or 2?). If you use UUID's, some of the chars and/or encodings can cause serious problems, that take many hours/days to fix in a Tex file.

itprojects:
Even if you can reduce this code more, you probably shouldn't. Easy debugging and quicker bug reporting have great value in the long run.

src/main.js
itprojects: See comments later in thread.

johnfactotum:
It's true that concise code is sometimes too clever and hard to understand, but in this case I think it would be much clearer if you can reduced the code a bit. And like I've said, should use template strings instead of `format`, and use `const` instead of `var` or `let`. Also I meant that you should add an explanation to the `ref${Math.round(Math.random() * 10000)}` in the code itself.

itprojects:
@johnfactotum In your code there's a bug, how do you advise we solve: `JS ERROR: SyntaxError: missing } in template string`

johnfactotum:
Line 21 in my example should probably be
```javascript
note = {${_('Quote: ') + esc(text) + (note ? _(" Note: ") + esc(note) : '')}}
```

itprojects:
Yes, indeed it was an extra bracket.

itprojects:
@johnfactotum Code should be ready to merge now.

src/main.js
```diff
@@ -2742,6 +2768,7 @@ class BookViewerWindow {
             const combo = new ComboBoxBox(_('Choose export format:'), null, [
                 ['html', _('HTML')],
                 ['txt', _('Plain Text')],
+                ['bib', _('BibTex')],
```
johnfactotum: One nitpick: should be "BibTeX", not "BibTex".

src/main.js
```diff
@@ -376,6 +376,32 @@ ${note}
...
+    // Math functions needed to avoid Tex problems
+
+    return annotations.map(({ text, note }, i) =>
+`@book{ref${Math.round(Math.random() * 10000)}:${i + 1},
+    author = {${esc(metadata.creator) || 'unknown'}},
+    publisher = {${esc(metadata.publisher) || 'unknown'}},
+    year = {${metadata.pubdate ? metadata.pubdate.slice(0, 4) : 'unknown'}},
+    note = {${_('Quote: ') + esc(text) + (note ? _(" Note: ") + esc(note) : '')}}
+},
+`).join('')
```
johnfactotum: Maybe put the comma and new line in `join` so there won't be a trailing comma at the end? Although I'm not familiar with BibTeX so I don't know if that would be good.

src/main.js
```diff
@@ -376,6 +376,32 @@ ${note}
...
+        : c).join('')
+
+    // Math functions needed to avoid Tex problems
+
+    return annotations.map(({ text, note }, i) =>
+`@book{ref${Math.round(Math.random() * 10000)}:${i + 1},
+    author = {${esc(metadata.creator) || 'unknown'}},
+    publisher = {${esc(metadata.publisher) || 'unknown'}},
+    year = {${metadata.pubdate ? metadata.pubdate.slice(0, 4) : 'unknown'}},
+    note = {${_('Quote: ') + esc(text) + (note ? _(" Note: ") + esc(note) : '')}}
```
johnfactotum: Should there be a new line or some other kind of delineation between the quote and the note? Should the quote be put inside quotation marks?

itprojects: That can be ignored.

src/main.js
itprojects: BibTeX can ignore this, and probably will.

itprojects:
@johnfactotum Although the new code produces a BibTeX, there is now a bug in it. You're supposed to use one reference for the book (ref123) and then cite with `\cite{ref123:1}` and `\cite{ref123:2}`, etc. It will be really confusing to any BibTeX user to have different book reference.

itprojects:
@johnfactotum Have you checked out the code?

johnfactotum:
>  Although the new code produces a BibTeX, there is now a bug in it. You're supposed to use one reference for the book (ref123) and then cite with `\cite{ref123:1}` and `\cite{ref123:2}`, etc. It will be really confusing to any BibTeX user to have different book reference.

I'm not sure what you mean as I have no experience using BibTeX. But I assume that this is still a problem?

I'm thinking that perhaps it would be better for this to be an external script, rather than something built into Foliate itself. Now that the book's metadata is included in the JSON file, it should be even easier than before to do just that.

itprojects:
> I'm not sure what you mean as I have no experience using BibTeX. But I assume that this is still a problem?

@johnfactotum No, that's solved in the latest commit to the merge/pull request.

> I'm thinking that perhaps it would be better for this to be an external script

That would actually take much more work from the user, and inflict a certain degree of animosity against Foliate developers, "that could have just put it in the app".

On the other hand - the final version is mature, optimized and will make Foliate an indispensable tool for any researcher - student or pro. That's what will distinguish Foliate from the competition.

johnfactotum:
Okay. I will merge this now.

PR #146, opened 2019-07-22
Null hrefs

itprojects:
Href is too often null, works much better without this.


itprojects:
Also worth evaluating:

`if (el.nodeName === 'A') {`
    `const parent = el.parentNode`
    `if (parent && (parent.nodeName === 'P' || parent.nodeName === 'DIV')) {`
        `el = parent`
    `}`
`}`

PR #172, opened 2019-09-17
Added copy to selection options

joyfulmantis:
This is a critical use case for me, as it allows me to  have a third party dictionary that is monitoring the keyboard and that will present the definition upon word selection


johnfactotum:
Thanks!

joyfulmantis:
Thank you!

PR #179, opened 2019-09-28
Fix glyph clipping in WebKit. Fixes #177

### LINKED ISSUE #177: Text at margin edge gets cut off
Some font glyphs can appear outside the margin, such as the letter `f`, and it's more visible with italic text. It happens on both side of a page, although it's most visible on the right side.

The issue can be seen on the Foliate screenshots (the word 'of').
screenshot

Here is a screenshot of mine. Increased the font to make it more visible.
Screenshot from 2019-09-27 08-48-38

I found Foliate the other day and started testing my ebooks with it. I think it's the best ereader currently on the Linux desktop. Keep up the good work!

daniel-j:
Before:
Screenshot from 2019-09-28 22-23-21

After:
Screenshot from 2019-09-28 22-22-38

Does not fix glyph clipping in vertical writing modes. Shouldn't affect anything else. Got the CSS property from Apple Books' CSS.

Testing EPUB: https://djazz.se/files/testcase.epub

I have submitted an identical PR to epub.js: https://github.com/futurepress/epub.js/pull/985

### DIFF
--- a/src/assets/epub.js
+++ b/src/assets/epub.js
@@ -5280,6 +5280,10 @@ var Contents = function () {
 
 			this.css(COLUMN_GAP, gap + "px");
 			this.css(COLUMN_WIDTH, columnWidth + "px");
+
+			// Fix glyph clipping in WebKit
+			// https://github.com/futurepress/epub.js/issues/983
+			this.css("-webkit-line-box-contain", "block glyphs replaced");
 		}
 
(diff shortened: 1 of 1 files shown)


johnfactotum:
Thanks a lot for taking the time and effort to investigate this issue! The fix seems to work well.

PR #198, opened 2019-11-01
update russian localization

zluca:


johnfactotum:
Thanks for this PR. The master branch contains a work-in-progress rewrite, and the strings are not stable yet, and I have not updated the translation files, so they all still contain the old strings from the 1.x branch. I'll merge this for now, but bear in mind that a lot of the strings might not be relevant anymore for 2.0.

Issue #199, opened 2019-11-01
[Feature Request] A new icon


itprojects:
Foliate is undergoing changes. A new icon, perhaps?

johnfactotum:
I will not be making a new icon just for the sake of change. But pull requests are always welcome.

PR #200, opened 2019-11-03
A new icon

itprojects:
License is GPL3 or later.

### DIFF
--- a/data/com.github.johnfactotum.Foliate.svg
+++ b/data/com.github.johnfactotum.Foliate.svg
@@ -6,57 +6,33 @@
    xmlns:svg="http://www.w3.org/2000/svg"
    xmlns="http://www.w3.org/2000/svg"
    xmlns:xlink="http://www.w3.org/1999/xlink"
-   xmlns:sodipodi="http://sodipodi.sourceforge.net/DTD/sodipodi-0.dtd"
-   xmlns:inkscape="http://www.inkscape.org/namespaces/inkscape"
-   id="svg8"
-   version="1.1"
+   width="64"
+   height="64"
    viewBox="0 0 110 73.000002"
-   height="128"
-   width="128"
-   sodipodi:docname="4.svg"
-   inkscape:version="0.92.4 (5da689c313, 2019-01-14)">
-  <sodipodi:namedview
-     pagecolor="#ffffff"
-     bordercolor="#666666"
-     borderopacity="1"
-     objecttolerance="10"
-     gridtolerance="10"
-     guidetolerance="10"
-     inkscape:pageopacity="0"
-     inkscape:pageshadow="2"
-     inkscape:window-width="1366"
-     inkscape:window-height="731"
-     id="namedview191"
-     showgrid="false"
-     inkscape:zoom="2.9610096"
-     inkscape:cx="90.15223"
-     inkscape:cy="85.670388"
-     inkscape:window-x="0"
-     inkscape:window-y="0"
-     inkscape:window-maximized="1"
-     inkscape:current-layer="layer1" />
+   version="1.1"
+   id="svg8">
   <defs
      id="defs2">
     <linearGradient
        id="linearGradient4617">
       <stop
-         id="stop4613"
+         style="stop-color:#f9f9f9;stop-opacity:1;"
          offset="0"
-         style="stop-color:#f9f9f9;stop-opacity:1;" />
+         id="stop4613" />
       <stop
-         id="stop4615"
+         style="stop-color:#f9f9f9;stop-opacity:0;"
          offset="1"
-         style="stop-color:#f9f9f9;stop-opacity:0;" />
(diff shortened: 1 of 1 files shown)


johnfactotum:
@bertob Maybe you can provide us with your professional opinion?

bertob:
Not a good direction IMO. It's not recognizable at all at 64 and 32 pixels (which are the main sizes you should test for), and it also doesn't really follow the guidelines with regard to perspective/style/color. The current icon is much nicer in both areas.

I don't want to discourage you from giving it another try though. Have a look at https://developer.gnome.org/hig/stable/icons-and-artwork.html.en and https://gitlab.gnome.org/Teams/Design/HIG-app-icons/wikis/App-Icon-Design-Workflow for some guidance on how to make icons that fit in with the rest of the platform.

itprojects:
The icon has been modified to make it more visible, in Update 1.

@bertob If you want to collaborate, perhaps make a version or two, that would be welcome.

bertob:
@itprojects I'm not sure what specific problems you're trying to address with this redesign. Perhaps that should be discussed in an issue first?

In general it's usually good to document what issues you see with the current icon, and discuss it with the maintainer and other community members going off and redesigning an app's icon. It's the identity of an app, and changing it is kind of a big deal :)

itprojects:
@bertob All that is true, and your feedback was taken onboard.

What problems are being addressed with the redesign?

The Foliate icon does not fit in with the typical theme environment, where the app runs (gnome-shell). The icons appears to be closer to a web-app icon. This does not represent in the best way possible an excellent desktop app.

The proposed new icon (see latest changes) looks more like the Foliate app window.

Community Discussions: See https://github.com/johnfactotum/foliate/issues/199

@johnfactotum Is the current work on a refreshed icon to continue?

johnfactotum:
In what way does the current icon appear "closer to a web-app icon"? I've tried to follow GNOME's general guidelines when creating the icon, and it doesn't look out of place from other GNOME icons:
image

The proposed new icon (using fc8f65d, because Icon Preview refuses to open 25aaf36), for comparison:
image

In particular, I'm not sure why you've introduced the red ribbon? The main idea of the current icon is a leaf that represents both the name of the app and its function (a leaf used as a bookmark). Splitting it into a red ribbon and small leaves (or an "F") makes the icon visually more complex and less recognizable.

I think a better direction might be to keep the overall concept of the current icon, as well as the cyan-ish primary color (which I feel is quite essential to to the identify of the app), but maybe improve on the shape or shading of the book or the leaf.

So I will close this PR for now. If you do decide to keep working on the icon, maybe you can post your ideas/sketches in the issue comments, so people can compare and provide feedback on the various proposals, before opening a new PR?

itprojects:
> In what way does the current icon appear "closer to a web-app icon"?

@johnfactotum  Cyan is very unusual. In nature, leaves are green / dark green / yellow / orange / red.

johnfactotum:
The dark cyan is more of a "stylized green", if you will. I don't see how it makes the icon "closer to a web-app icon"?

But if you must insist on a less blue color, perhaps we can try the greens from [GNOME's palette](https://developer.gnome.org/hig/stable/icon-design.html).

Maybe we should change the current icon's book covers to a darker one or gray color, which meshes with the gray of the pages better.
image

I kind of like this autumn leaf, too:
image

bertob:
Just to clarify: for third party apps the guidelines don't discourage people from using custom, non-palette colors. The palette is usually a good starting point, but unusual colors can be an important part of the identity for non-core apps. I don't see any issues with the original cyan, for example.

johnfactotum:
Thanks for the clarification. I do like the vibrant colors in the palette better than the darker colors in @itprojects's proposed icon. And the orange leaf is kind of interesting. In the olden days it would be fun to do a changing icon based on season à la VLC's Santa hat icon.

itprojects:
Orange and green, for the leaf, and a dark book cover would be great.

Some casual suggestions below:



Since the leaf is a bookmark ribbon, shouldn't it be more like this:



Foliate and  Gnome-books icons shun any real text in the icon, strange for a text read app. One cannot say that the icons are identical, but the same inspiration is at play.

johnfactotum:
The leaf is just a leaf pressed inside the book as a bookmark; it's not a leaf-shaped ribbon. I think the large border radius looks nicer and fits with other GNOME icons, and the depth make the icon more distinct and book-like.

itprojects:
The depth seems the same as the one in the Gnome-books icon.

Books rarely have rounded corners (unless in the case of journals / diaries, or herbariums).

Consider this [modification](https://postimg.cc/svybHvWr).

johnfactotum:
The book in Gnome Books' icon is much thicker than the one in the last icon in your previous comment, and it also has the same rounded corners? I'm not quite sure what you're getting at?

itprojects:
It was only a thought.

Issue #224, opened 2020-01-02
Swipe to turn page skips 4 pages at a time


tokyovigilante:
Using 1.53 (but also occurs with git build) on FC31, ThinkPad X1 Yoga. In landscape with two pages visible, swiping to turn page skips ahead twice (ie 4 pages ahead). It can be avoided by using a tiny slide/swipe but this is very finicky.

johnfactotum:
I assume you're referring to touchscreen swipes? Currently the code only handles the wheel event, and uses debouncing so that a single scrolling action would only turn the pages once.

https://github.com/johnfactotum/foliate/blob/22cfa61d31d1d31755c66bfa4eb962907557ba8b/src/assets/epub-viewer.js#L622-L635

https://github.com/johnfactotum/foliate/blob/22cfa61d31d1d31755c66bfa4eb962907557ba8b/src/assets/utils.js#L16-L28

I've only tested this with a laptop touchpad as I don't have any touchscreen device to test with. Actually I didn't expect the wheel event to be triggered on a touchscreen at all, and it's not clear how exactly WebKitGTK handles touch devices.

touchscreenuser:
Worked on Arch Linux Surface Go with master branch without issues. Which distro / version of WebKit2GTK are you using?

jthornhill:
Just a note that I also observe inconsistent swipe page turns. If I am very precise and careful to make small swipes I can seem to mostly get single page turns as @johnfactotum mentioned, but not totally reliably.

I'm running Foliate on a Dell XPS 9360 on Arch Linux using GNOME. The versions are:

extra/webkit2gtk 2.26.2-2 [installed] 
community/foliate 1.5.3-1 [installed] 

(As an aside even with this issue Foliate works relatively well on my touchscreen, when compared to other Linux ebook reading software!)

touchscreenuser:
I use Cinnamon on Arch Linux with Foliate 2.0.0 from AUR repository.

nvllsvm:
I've disabled the mouse wheel in epub-viewer.js so I can solely use tap to turn. https://github.com/nvllsvm/foliate/commit/ef948d64d8eafd93035b64309901e05eb08ebbd5

Similar to the "Turn page on tap" option, perhaps a "Turn page on wheel" option could be added to settings. Thoughts?

johnfactotum:
@nvllsvm See #555.

PR #232, opened 2020-01-15
[WIP] Add touch event handler

tokyovigilante:
Regarding #224, it seems that touch events are being sent as scroll wheel events. Nothing wrong with this specifically, although it would be better if WebKitGTK supported W3C pointer events, but this seems to be blocked on Apple not supporting it outside iOS Safari.

Anyway, WebKitGTK does support the older Touch Events API, so I've added this basic handler to epub.js to handle touch swipes. Annoyingly, the event handler uses a debounce to prevent scrolling multiple pages at a time seemingly, so I've had to add the `touchStarted` boolean to the `rendition` object. Apparently otherwise the `preventDefault()` calls are enough to block double-actions from both wheel and touch events. I didn't invent JS...

Annoyingly, this only seems to work currently with forward swipes. The `touchend` event is not called on a swipe backward, and I get the old behaviour of two page scrolling. One step forward and two steps back it seems. I'm not sure why this is as even the `rendition.prev()` function doesn't seem to be called, suggesting it isn't scroll wheel events either, maybe something to do with WebKitGTK enabling back and forward navigation on swipe in some cases (although this is opt-in and doesn't seem to be enabled).

This works well enough for me on forward swipe so I thought I'd post it up for comment. I don't know if it would be accepted upstream because ideally pointer events would be targeted for other browsers, but the code is modified from [upstream](https://github.com/futurepress/epub.js/wiki/Tips-and-Tricks-%28v0.3%29#swipe-to-turn-pages-with-touchevents), so once we get it working well I can try to push it there.

### DIFF
--- a/src/assets/epub-viewer.js
+++ b/src/assets/epub-viewer.js
@@ -618,9 +618,40 @@ const setupRendition = () => {
     rendition.on('keydown', handleKeydown)
     document.addEventListener('keydown', handleKeydown, false)
 
+    //Enable swipe gesture to flip a page
+    let start;
+    let end;
+
+    rendition.touchStarted = false;
+
+    document.addEventListener('touchstart', (event) => {
+        start = event.changedTouches[0];
+        rendition.touchStarted = true;
+        event.preventDefault();
+    });
+
+    document.addEventListener('touchend', (event) => {
+        rendition.touchStarted = false;
+        if (!paginated) { return };
+        event.preventDefault();
+        end = event.changedTouches[0];
+        const elBook = document.querySelector('#viewer'); //Parent div, which contains the #area div
+        if (elBook) {
+            const bound = elBook.getBoundingClientRect();
+            const hr = (end.screenX - start.screenX) / bound.width;
+            const vr = Math.abs((end.screenY - start.screenY) / bound.height);
+            if (hr > 0.25 && vr < 0.1) return rendition.prev();
+            if (hr < -0.25 && vr < 0.1) return rendition.next();
+        }
+    });
+
     if (paginated) {
         // scroll through pages
         const onwheel = debounce(event => {
+            if (rendition.touchStarted) {
+                event.preventDefault();
+                return;
+            }
             if (getWindowIsZoomed()) return
             const { deltaX, deltaY } = event
             if (Math.abs(deltaX) > Math.abs(deltaY)) {


johnfactotum:
Thanks for working on this. I think we should investigate what exactly is causing it to skip pages. Is it indeed a bug with the `onwheel` function? If not, perhaps one can try [Epub.js's example pages](https://futurepress.github.io/epub.js/examples/) and see if the issue occur there as well.

tokyovigilante:
Hi John, thanks for the look. Sorry I didn't realise until reviewing the commit history today that epub-viewer.js is your code and not upstream. The epub.js viewer doesn't seem to have any mouse wheel or touch handling so I'm not sure they'll help us. 

The other thing I noticed is that the upstream viewer seems to have significantly better page-turning performance, and there is at least one un-necessary paint (and probably two) in profiling during page turns. I wonder if this can be improved? The trade off of course is that Foliate is SO. MUCH. PRETTIER. ;)

jhakonen:
I stumbled on this PR as I was trying to figure out why Foliate was infrequently flipping 2 pages at same time when I was swiping the reader. It seems to me that this PR partially fixes that bad behaviour for me.

By adding some `console.log()` calls to the event handlers, I noticed that the `touchstart` event only triggers if the swipe starts at left or right edge of Foliate's window. The event is not triggered if the swipe starts from window's center.

Even though the touch event handlers are attached to the top level `document`, a touch event starting within the viewer element (perhaps because it is a iframe?) doesn't seem to be getting caught by the event handlers. 

With that observation, I created another set of `touchstart` and `touchend` handlers, with exactly the same content as the top level handlers, but attached them to `contents.document` instead. This seems to have fixed the swiping so that I'm not getting extra page flips any more when the swipe doesn't start from the edge.

Is checking of the `vr` check necessary? I seemed to have some difficulties getting my swipes horizontal enough for the page flip to trigger. Removing the vertical check made the swiping easier for me.

jhakonen:
Here's my changes: https://gist.github.com/jhakonen/27a06112087a99512ceb3b952f52cc21

johnfactotum:
@jhakonen Can you make a PR for your changes based on the latest master? `src/assets/epub-viewer.js` was moved to `src/web/epub-viewer.js`.

jhakonen:
It turned out that the way the swipe is detected was still somewhat hard to use. You have to drag your finger across the screen for  25% width of the screen for the swipe to be detected.
There is however a class in Gtk that already implements swipe gesture detection: [Gtk.GestureSwipe](https://gjs-docs.gnome.org/gtk30~3.24.22/gtk.gestureswipe)
With that class, instead of dragging your finger across the screen, you only need the swipe your finger fast enough for the gesture to be detected. This way you do not need move your hand at all, as just a fast swipe of a finger is enough.
I've rewritten the touch handling using Gtk.GestureSwipe and I would like to make a PR out of that instead.

johnfactotum:
Closing this now that #591 is merged

PR #237, opened 2020-01-28
update russian translation

zluca:


po/ru_RU.po
```diff
@@ -55,7 +55,7 @@ msgstr "Открыть файл"
 
 #: src/main.js:103
 msgid "translator-credits"
-msgstr ""
+msgstr "переводчики"
```
johnfactotum: "translator-credits" should be the list of translators, separated by newlines, like
```
Schwonder Reismus <schw0reismus@protonmail.com>
Artem Polishchuk <ego.cordatus@gmail.com>
zluka <https://github.com/zluca>
```

zluca: Ok. I just deleted it.

PR #243, opened 2020-02-11
Updated Dutch translation and merged nl_NL.po into generic nl.po

Vistaus:
:-)


johnfactotum:
Thanks! Should the `LINGUAS` file also be updated?

Vistaus:
You're right, I forgot to do that! Thanks for noticing! I just updated LINGUAS.

PR #244, opened 2020-02-14
Some translation and adaptations.

msmafra:
Translated non translated strings and made some translation adaptations.


po/pt_BR.po
```diff
@@ -46,15 +46,15 @@ msgstr "Todos os arquivos"
...
 
 #: src/main.js:89
 msgid "Open File"
-msgstr "Abrir arquivo"
+msgstr "Abrir Arquivo"
 
 #: src/main.js:103
 msgid "translator-credits"
-msgstr ""
+msgstr "Créditos de tradutor"
```
johnfactotum: "translator-credits" should be the list of translators, separated by newlines, not the literal translation of "translator-credits".

Issue #268, opened 2020-04-06
Minor icon update?


bertob:
The current icon is nice overall, but the "fold" at the bottom is a bit odd. I was playing with this book shape with a proper spine, and I feel like it could work:

![image](https://user-images.githubusercontent.com/1908896/78602153-3b8f7780-7856-11ea-998c-919acebe3e74.png)

If you're interested in pursuing this direction I'd be happy to iterate on this and try some variations.

johnfactotum:
Yeah, I think I like this direction. It would give it a much more distinct outline, instead of just a dull rectangle. Although the steep and abrupt angle change of the fold kind of makes it look like the book is melting -- maybe that's just because I'm used to the old icon.

bertob:
If you want I can do an MR so you can try it for real, and we can iterate from there.

johnfactotum:
Sure!

johnfactotum:
Thanks a lot for your work!

PR #279, opened 2020-04-08
Icon: new book shape with more realistic spine

bertob:
As discussed in #268 

image


johnfactotum:
Actually, it doesn't look so weird now, after getting used to it for a while. Do you have any further variations on this idea? Maybe we can get this into v2.1.0.

bertob:
Tried a more realistic leaf, and some minor shading tweaks on the book:

image

johnfactotum:
Looks good to me, honestly. The new shading looks much better. I think I'm going to go ahead and merge this.

Issue #301, opened 2020-04-14
[Feature Request] Support StarDict offline dictionaries


habenamare:
The current support for offline dictionaries with `dictd` works well. I think we should extend this feature by adding support for dictionaries in the **StarDict** format (which are widely available).

habenamare:
I have created PR #302, which implements this feature using the `sdcv` program.

PR #302, opened 2020-04-14
add support for StarDict dictionaries

habenamare:
This PR implements #301.

The implementation is very similar (almost the same) to the `dictd` implementation. Here, the `sdcv` program is used instead of the `dict` program.

Tested on `sdcv` version 0.5.1


johnfactotum:
Looks good to me. Thanks!

Issue #254, opened 2020-03-30
Support CBZs ($50 Bounty)


anthonyajumelles:
Hello there,
I am deeply appreciative of all the work that has gone into making Foliate the premier eBook reader. The ground up rewrite looks stellar. 

My only request is for CBZ support to be considered. I know Komikku is available for manga however I would much prefer to be able to utilize my collection of legally purchased CBZs.  

KOReader is also an alternative but the UI is nowhere near as good as Foliate.

Normally I try not to request new features, however both EPUBs and CBZs are zips and the main work for this would be displaying images in sorted order, something that I wouldn't say is trivial but that fits within the scope of the applications current features.

The only extra features I can conceive of are cropping margins, inverting images for night mode, and converting white space to background color. Baseline of just fitting the image to display width would do the trick.

$50 USD to start: https://www.bountysource.com/issues/90789000-request-support-cbz-comic-books-manga-w-bounty

habenamare:
I have created PR #314 which implements this issue. I would appreciate any feedback, especially on how the pages are displayed.

habenamare:
This issue is not closed at Bountysource. @johnfactotum Can you check the integration? Maybe using [Bountysource Tools](https://www.bountysource.com/tools)?

johnfactotum:
Okay, I've installed the plugin. I've also sent a request to claim the foliate "team", which should be "processed within 2 business days."

PR #314, opened 2020-04-22
Add support for Comic Book container formats (CBZ, CBR, ...)

habenamare:
This PR implements #254.

> These formats are compressed archives that contain images in alphabetical order and the images represent pages of a comic book. The difference between the formats is their compression algorithm.

The support for the comic book container formats is implemented by generating a [WebPub Manifest](https://github.com/readium/webpub-manifest) from the images inside the archives.

**Tasks**
- [x] Add `CBZ` support using `JSZip`

- [x] Embed the [libarchivejs]() package

- [x] Add a separate page for displaying CB files because `libarchivejs` needs a custom **Content Security Policy**

- [x] Add `CBR` support using `libarchivejs`
   > Some `RAR` files can not be extracted because of a bug inside the [libarchive](https://github.com/libarchive/libarchive) library. For more info, see the [open issue](https://github.com/libarchive/libarchive/issues/373) for the bug.

- [x] Add `CB7` support using `libarchivejs`

- [x] Add `CBT` support using `libarchivejs`

- [x] When extracting archives, filter by file type. The filtering file types should be: `image/jpeg`, `image/png`, `image/gif`, `image/bmp`, `image/webp`

- [x] Optimize page display based on current layout
  - **Two-Page View** for the *Automatic* layout
  - **Fit Page** for the *Single Column* layout
  - **Fit Width** for the *Scrolled* & *Continuous* layouts

    > **Two-Page View**, **Fit Page** and **Fit Width** are affected by the `Width` setting. The value of the `Width` setting can significantly vary the window's width from the actual book displaying width.


johnfactotum:
@habenamare I've figured out a way to convert files for Epub.js's consumption without intermediate files by using webpub manifests and blob urls -- see my implementation of plain text support.

habenamare:
@johnfactotum If I understand correctly, this means that these comic book formats should be converted to `webpub` instead of `opf`.  And for the images inside, they should be represented using Blob URLs instead of being stored in a temporary directory.

If my understanding is correct,
1. Should `javascript` be used for extracting the archives?
2. How will this affect memory usage? These comic book formats can sometimes get very large and extracting and representing them in memory might increase memory usage.

johnfactotum:
> If I understand correctly, [...]

Yes. Also note that even when unpacking to a temp directory, you can still use the WebPub format, which is slightly easier to work with than EPUB.

> Is javascript going to be used for extracting the archives?

Since Epub.js already uses JSZip, I think it makes sense to reuse that. Not sure what's available for other archiving formats, though.

> How will this affect memory usage? These comic book formats are sometimes very large and extracting and representing them in memory might increase memory usage.

That's a good point. I guess it's a trade-off, because disk usage can be very expensive, too. I suppose unpacking might turn out to be a better fit in some situations, but I just want to share this alternative approach, which I think is simpler and requires no temp files.

habenamare:
Thanks for the approach. You're right, the WebPub format is much more simpler to generate than EPUB.
 
> Since Epub.js already uses JSZip, I think it makes sense to reuse that. Not sure what's available for other archiving formats, though.

At first, I was going to use JSZip. Then, I decided to implement extracting the archives in Python because I thought it would be easier to support many archive formats. But, now I think I should stick with JSZip (since it's already included) and I have found a package that handles other formats.

The package is [libarchivejs](https://github.com/nika-begiashvili/libarchivejs) and it is a WebAssembly port of the [libarchive](https://github.com/libarchive/libarchive) C library.

habenamare:
@johnfactotum I have embeded the `libarchivejs` package and it is working very well. But, there is something with the library. Since it uses WebAssemby, it needs to use `eval`. And for the library to work the **Content Security Policy** of `epub-viewer.html` needs to be changed. What do you think?


```html
<meta http-equiv="Content-Security-Policy" content="...; script-src 'self';...">
```
TO

```html
<meta http-equiv="Content-Security-Policy" content="...; script-src 'self' 'unsafe-eval';...">

```

johnfactotum:
I'm not really sure. Did some testing, and it seems that with `unsafe-eval` the book itself won't be able to execute anything. However, if we ever mutate the document in any way from untrusted input (currently, at least I believe, we don't), then it'll be a problem.

IMO the risk isn't really _that_ high. But if there's really no way around `unsafe-eval`, maybe it's best to use a separate page specifically for opening CB files? Or one can simply load CB files without CSP, which should be safe as long as the file names are sanitized.

habenamare:
Good points. For now, I have created a separate page for displaying CB files. And, you're right about sanitizing user input, which in this case is name of the comic book archive and name of the files inside the archive.

habenamare:
> Good points. For now, I have created a separate page for displaying CB files. And, you're right about sanitizing user input, which in this case is name of the comic book archive and name of the files inside the archive.

After doing some research, I don't think sanitizing the file names is necessary. This is because the `eval` function is not used with user input, in fact it is not used anywhere and it is only used to make some part of a library work.

src/assets/webpub.js
```diff
@@ -173,3 +173,83 @@ const processFB2 = doc => {
...
+            return {
+                href: pageURL,
+                type: 'text/html',
+                title: page.title
+            }
+        })
+
+    return {
+        metadata: {
+            title: cbArchiveName
```
johnfactotum, line 248: Add `layout: 'pre-paginated'` to render it as a fixed-layout EPUB.

habenamare: I think I have found a better way to get predictable results. Let me know if you have any ideas.

johnfactotum: Nice. But I think `layout: 'pre-paginated'` and `spread` should be turned on for the "Automatic" layout?

habenamare: Yes, you're right. The **Automatic** and **Single Column** layouts shouldn't have the same effect in the first place. So, this means,
  - If the selected layout is **Automatic**,
    ```js
    return {
        metadata: {
            title: cbArchiveName,
            layout: 'pre-paginated',
        },
        ...
    }
    ```
  - Else,
    ```js
    return {
        metadata: {
            title: cbArchiveName
        },
        ...
    }
    ```

johnfactotum: You can use `layout: 'reflowable'` for the other layouts.

habenamare: Ok, I think `reflowable` was the default option or something but it is better to explicit.

johnfactotum: > But, not disabling `spread` gives unpleasant results.

What do you mean by "unpleasant results"? The whole point of `layout: 'pre-paginated'` is so that it would show a two-page view, so `spread` should definitely be enabled. If you mean sizing or alignment issues, I think your original flexbox stylesheet works well with this mode.

habenamare: Yes, I meant sizing and alignment issues when I said "unpleasant results". But, you're right everything looks good with the previous flexbox-based stylesheet. Thanks for making things clear.

src/assets/webpub.js
```diff
@@ -173,3 +173,173 @@ const processFB2 = doc => {
...
+    switch (inputType) {
+        case 'cbz': files = await unpackZipArchive(uri); break
+        case 'cbr': files = await unpackArchive(uri, inputType); break
+        case 'cb7': files = await unpackArchive(uri, inputType); break
+        case 'cbt': files = await unpackArchive(uri, inputType); break
+    }
+
+    const sectionLinkObjects = files.filter(file =>
+            ['jpeg', 'png', 'gif', 'bmp', 'webp'].includes(file.type))
+        .map(image => {
```
johnfactotum, line 305: Currently the images don't appear to be sorted correctly for some books. See https://github.com/nika-begiashvili/libarchivejs/issues/22. Maybe add a `.sort((a, b) => a.name.localeCompare(b.name))` here?

johnfactotum:
@habenamare Is this ready for merging?

@anthonyajumelles Have you tried this out? Any comments?

habenamare:
@johnfactotum Yes, it's ready.

anthonyajumelles:
@johnfactotum I have zero complaints! I was testing @habenamare's branch last night with no issues.

johnfactotum:
Great! Thanks for the great work!

habenamare:
@johnfactotum No problem, Thanks for your work on this great eBook reader!

Issue #327, opened 2020-05-02
Support removable media in Foliate snap


D15C0DE:
Please add the removable-media plug to snapcraft.yaml so Foliate will be able to read files from /media and /mnt.

apps:
  foliate:
    plugs:
      - removable-media

Thanks for this great epub reader!

Issue #324, opened 2020-04-30
Snap cannot run TTS


molten-steel:
On the Foliate snap (v2.1.1) trying to run text to speech returns a warning `GLib.SpawnError: Failed to execute child process “espeak-ng” (No such file or directory)`

johnfactotum:
Paging @galgalesh.

galgalesh:
Thanks for this report @molten-steel 

Foliate supports TTS using many different programs such as `espeak-ng` and `festival`. I can add one or more of them to the snap, but I'd like to make an informed decision about what to add specifically. @molten-steel Do you have any experience with this TTS software? `festival` sounds much better than `espeak-ng` to me, for example. Do you have experience with some of the voices?

Also, what language do you want to use this for?

molten-steel:
Since the default voice isn't great, I used to run it on flatpak with mbrola using the command `espeak-ng -v mb-en1 -s 210` which gives me a better voice and allows me to set the speed.

The `male;awb` voice on flite seems good, but I'm having trouble finding a list of voices on the full version of festival, so I'm not sure if that's an option. I'd be fine with whatever festival voice you think is best, but is there a way to allow changing the speed if festival is packaged in the snap?

molten-steel:
If I'm reading [this](http://www.cstr.ed.ac.uk/projects/festival/manual/festival_24.html) correctly, it has my favorite mbrola voice as `voice_en1_mbrola`, so I'd recommend that one.

PR #329, opened 2020-05-03
Snap: Support TTS and removable-media

galgalesh:
Fixes https://github.com/johnfactotum/foliate/issues/327 and https://github.com/johnfactotum/foliate/issues/324

This PR adds the `removable-media` interface to allow Foliate to open files on USB sticks. This PR also adds the necessary packages for TTS in the snap. It also changes the default TTS command of the snap to `espeak-ng -v mb-en1` so that TTS is enabled by default. It think it makes sense to provide a default command for the snap since we are certain `espeak-ng` will always be available, and since users can't really choose the command themselves because it depends on what we actually packaged in the snap.

It's not ideal that users are restricted to the voices and synthesizers we ship in the snap, so I also started [a discussion on the Snapcraft forum for better support for TTS in snaps](https://forum.snapcraft.io/t/text-to-speech-support-in-snapd/17099). It would be better to either use Speech Dispatcher of the host system, or to package Speed Dispatcher itself as a snap.

### DIFF
--- a/build-aux/snap/snapcraft.yaml
+++ b/build-aux/snap/snapcraft.yaml
@@ -15,6 +15,7 @@ apps:
       - opengl
       - network
       - audio-playback
+      - removable-media
     slots:
       - dbus-daemon
     common-id: com.github.johnfactotum.Foliate
--- a/build-aux/snap/snapcraft.yaml
+++ b/build-aux/snap/snapcraft.yaml
@@ -14,6 +14,7 @@ apps:
       - home
       - opengl
       - network
+      - audio-playback
     slots:
       - dbus-daemon
     common-id: com.github.johnfactotum.Foliate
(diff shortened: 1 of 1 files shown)


johnfactotum:
Thanks for the PR!

Launching TTS programs on the host is what I've done with the Flatpak package, with `flatpak-spawn`. Is there / could there be something similar for Snaps?

Also, apart from TTS, there's the same problem with offline dictionaries. Currently we support `dictd` and `sdcv` (support of the latter not released yet).

galgalesh:
> flatpak-spawn

In snaps, you would use classic confinement for this, but then the snap can run arbitrary host commands, so it's not really confined anymore. I'd like to avoid this because I think there is a big advantage to running apps in sandboxes, for both security and compatibility reasons.

Dictionaries and specifically `dictd` are similar to TTS in that their total size is large (200MB), but each user will only use a small subset. Users only need the packages for the languages they read. The difference between TTS and dictionaries, however, is that these dictionaries are rarely installed by default on the host, so it makes less sense to try to use a dictionary from the host system. Since the user already needs to install the dictionaries, we could easily just let the user install the dictionaries as snaps. Connecting two snaps so one snap can use the content is already possible. The Foliate snap already does this with the GNOME runtime.

The reason why I haven't done this for dictionaries, is because I don't want to maintain the snap with dictionaries. The advantage over the online dictionaries also seems small to me. It's a quite clunky UX that a user needs to manually install the dictionaries they want/need. The online dictionary just works out of the box.

galgalesh:
Now that I think of it, the ideal solution for the both the dictionaries and the TTS languages would be if there was a way for Foliate to request the installation of extra snaps. Foliate knows the language of the books and Foliate knows which dictionaries are supported, so if the dictionaries would be snapped, and Foliate could request snapd to install a dictionary, the UX would be really great.

Does this make sense? (As a long-long term solution)

johnfactotum:
Yeah, I think there should definitely be some kind of portal for this. Looking up words, text-to-speech, etc., are common and useful things with apps like browsers, document viewers, feed readers, email clients, etc. But currently each app must re-invent its own implementation, and a lot of them are missing these features. It'd be great if these can simply launch the desktop's search or dictionary app (which most desktops lack a decent one, but that's another problem), preferably in a more integrated way (like displaying the results in a popover, rather than launching a new window).

Other things that might require additional "content snaps" include things like hyphenation rules and spellchecking packages. I mention these, because many of these packages, similar to offline dictionaries or TTS voices (for the lesser languages, or better quality packages for the major langauges), are not packaged by distros, nor maintained as Flatpaks/Snaps. But for the user, it's much harder to workaround this for sandboxed apps. Admittedly, this might be more of a rare use case.

Anyway, for the moment I suppose this PR would provide the best out-of-the-box experience for Snap.

PR #331, opened 2020-05-05
fixed release date

yasharjit:
I noticed on flathub that the release version of foliate was 2.0.0 rather than the latest version of 2.1.1. Though when I install from flathub and launch the app I can see it is version 2.1.1.

On flathub
Screenshot from 2020-05-05 13-10-14

On using 'flatpak list' command in terminal
Screenshot from 2020-05-05 13-10-45

In the installed foliate app
Screenshot from 2020-05-05 13-11-27

So I have corrected the release date in appdata.xml file. Please take a look.

### DIFF
--- a/data/com.github.johnfactotum.Foliate.appdata.xml.in
+++ b/data/com.github.johnfactotum.Foliate.appdata.xml.in
@@ -43,7 +43,7 @@
         </screenshot>
     </screenshots>
     <releases>
-        <release version="2.1.1" date="2019-04-09">
+        <release version="2.1.1" date="2020-04-09">
             <description>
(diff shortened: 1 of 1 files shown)


johnfactotum:
Ah, so that's what broke the release info on flathub. I've been wondering about this. Would need to patch this in the flathub repo, too: https://github.com/flathub/com.github.johnfactotum.Foliate. Can you also make a PR there?

yasharjit:
I am not familiar with what changes I need to make in that repo. Can you please tell me what to change there?

johnfactotum:
Okay, I've pushed a fix myself. Thanks for the PR!

Issue #161, opened 2019-09-01
Join spread


olivierberten:
When using 2 pages view, both pages are left aligned while one would expect them to be joined in the middle, like epub.js does it. This is especially annoying for fixed layout epub with images across the spread.

johnfactotum:
Well, Epub.js renders fixed layout EPUBs like that by default, and I'm not sure how to fix it. As a workaround you can try resizing the window or adjusting the margins.

johnfactotum:
This has been fixed in Epub.js 0.3.86.

johnfactotum:
Reopening as the spread is not joined when there aren't `page-spread-*` properties

johnfactotum:
Fixed in the new renderer (https://github.com/johnfactotum/foliate/issues/962).

PR #334, opened 2020-05-05
Minor spacing change when displaying CB files

habenamare:
This is a minor change that adds some spacing when displaying CB files on the **Single Column** layout. This should have been included in PR #314, but I didn't notice it because of the frequent implementation changes.

@johnfactotum  @anthonyajumelles If you don't have any suggestions, this is ready to be merged. The change is minor and I don't think it will have any unexpected side effects.

## Single Column
The change makes top and bottom margins relatively equal.
  > The green background is added only to show the difference. It's not part of the change.

## Before
single-column-current

## Updated
single-column-new

### DIFF
--- a/src/assets/webpub.js
+++ b/src/assets/webpub.js
@@ -327,10 +327,10 @@ const webpubFromComicBookArchive = async (uri, inputType, layout) => {
         `
     }
     
+    const automaticScripts = async () => { return `` }
     const fitPageScripts = async () => { return `` }
     const fitWidthScripts = async () => { return `` }
     const continuousScripts = async () => { return `` }
-    const automaticScripts = async () => { return `` }
 
     let stylesheet;
     let scripts;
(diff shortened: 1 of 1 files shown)


anthonyajumelles:
It looks great! 

My only concern would be breaking dual page spreads but CBZs are responsible for anticipating that through inclusion of both pages in one image file.

johnfactotum:
I think I would prefer not having that padding in the automatic layout. The padding isn't really necessary, and it's nice to be able to have the pages join together seamlessly.

Plus, even without the padding, they don't join together if the window is too wide, which I'm not sure if it's possible to fix (see https://github.com/johnfactotum/foliate/issues/161). Maybe we can patch Epub.js and add extra classes or attributes so the stylesheet can detect whether it's currently displayed as a right or left page. That way we can have a seamless spread at all times.

habenamare:
@johnfactotum  I have reverted the changes for the **Automatic** layout as it doesn't seem necessary. Your other suggestion seems to be a better enhancement for the **Automatic** layout. I'll see if I can work something out. And, I think this PR should be merged since it is for a minor spacing change (now only for the **Single Column** layout).

Issue #118, opened 2019-07-10
Show commit hash in about dialog


itprojects:
It would be great for testing and bug reporting to know the commit string of the tested application build. Could this be added to the application "About" popup?

johnfactotum:
I've toyed with this idea before, but in the end it didn't seem very useful. Most people would probably be using the released versions. For people building from source, it should be easy enough to check which commit they're on.

itprojects:
One other idea would be to put the commit string in the command line:
`com.github.johnfactotum.Foliate --version`
to give:
`Foliate version 1.4.0 commit 5bf8f3ccc32603bf8daa57bcdd0806ca4bf085f0`

PR #337, opened 2020-05-08
snap: Updated to use the latest snapcraft extension and improve version

kenvandine:
Updated to use the latest snapcraft extension
Set a more meaningful version based on git tags

### DIFF
--- a/build-aux/snap/snapcraft.yaml
+++ b/build-aux/snap/snapcraft.yaml
@@ -9,7 +9,7 @@ confinement: strict
 apps:
   foliate:
     command: usr/bin/com.github.johnfactotum.Foliate
-    extensions: [gnome-3-28]
+    extensions: [gnome-3-34]
     plugs:
       - home
       - opengl
@@ -24,15 +24,15 @@ parts:
   foliate:
     plugin: meson
     source: .
+    override-pull: |
+      snapcraftctl pull
+      sed -i -e 's|@GJS@|/usr/bin/gjs|g' src/com.github.johnfactotum.Foliate.in
     override-build: |
       set -ex
       xmlstarlet edit -L -u "/schemalist/schema/key[@name='tts-command']/default" -v "'espeak-ng -v mb-en1'" $SNAPCRAFT_PART_BUILD/data/com.github.johnfactotum.Foliate.gschema.xml
       snapcraftctl build
     meson-parameters: [--prefix=/snap/foliate/current/usr]
     build-packages:
-      - libgjs-dev
-      - gettext
-      - libglu1-mesa
       - xmlstarlet  # for override-build
     organize:
       snap/foliate/current/usr: usr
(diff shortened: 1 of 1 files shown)


build-aux/snap/snapcraft.yaml
```diff
@@ -24,15 +24,17 @@ parts:
   foliate:
     plugin: meson
     source: .
+    override-pull: |
+      snapcraftctl pull
+      sed -i -e 's|@GJS@|/usr/bin/gjs|g' src/com.github.johnfactotum.Foliate.in
```
galgalesh, line 29: The sed hack is currently needed because meson sets the path to gjs using `$PATH` at build time. The gnome-3-34 extension puts the build snap’s path first in $PATH. So meson replaces @GJS@ with /snap/gnome-3-34-1804-sdk/current/usr/bin/gjs for the interpreter path. This is the correct path in the build VM, but will not work in the the actual snap.

We'd like to find a way to fix this without this `sed` patch.

build-aux/snap/snapcraft.yaml
```diff
@@ -24,15 +24,17 @@ parts:
...
+      snapcraftctl set-version $(git describe --tags --abbrev=10)
```
galgalesh, line 30: In the current version of the snap, we use the AppStream file to get the version. That works great for releases, but not for the daily builds in edge; they just get the same version as the previous stable release. I think it would be cleaner to have Meson generate the version in the AppStream file using git describe. Then we kan keep using the version from AppStream and all development builds will have a useful version string, instead of only the snap. However, I don't know if there is a good way to do that in Meson.

@johnfactotum What do you think about this? Should we try to get Meson to generate the version, or do it in the snap here?

build-aux/snap/snapcraft.yaml
```diff
@@ -51,11 +53,13 @@ slots:
...
   # This is a bind instead of a (faster) symlink because mbrola uses LD_PRELOAD
   # with libraries from this directory, which doesn't work with symlinks.
-  /usr/lib/x86_64-linux-gnu/mbrola:
-    bind: $SNAP/usr/lib/x86_64-linux-gnu/mbrola
+  /usr/lib/$SNAPCRAFT_ARCH_TRIPLET/mbrola:
+    bind: $SNAP/usr/lib/$SNAPCRAFT_ARCH_TRIPLET/mbrola
   /usr/share/mbrola:
     symlink: $SNAP/usr/share/mbrola
+  /usr/bin/gjs:
+    symlink: $SNAP/gnome-platform/usr/bin/gjs
```
galgalesh, line 65: This layout will be added by the `gnome-3-34` extension in the following release of snapcraft, so this won't be needed anymore in the future.

build-aux/snap/snapcraft.yaml
johnfactotum: I'm not really sure. In the past I've appended the commit hash to the version when built with the development option, and also added the `.devel` class, which, in Adwaita at least, adds a blue tint and cogs to the headerbar background (see #118). However I've not done this in the 2.0 rewrite, because it doesn't seem very useful.

In many GNOME apps, they have an option to build the development version separate from the stable version, and it's possible to install them side-by-side. Their development builds also use modified icons.

I think the proper way is to have Meson generate everything properly. Another option is to manually bump the version and add a development release to AppData after each stable release. This could be applied on its own or in conjunction with Meson adding the git hash.

But all these are pretty cumbersome to implement and personally I don't see a lot of advantage.

galgalesh: Got it!

PR #346, opened 2020-05-11
Add python/python3 to dependency, and PPA for Foliate

apandada1:
This would fix the last four lintian errors.

```
$ lintian com.github.johnfactotum.foliate_2.1.1_all.deb 
E: com.github.johnfactotum.foliate: no-copyright-file
W: com.github.johnfactotum.foliate: binary-without-manpage usr/bin/com.github.johnfactotum.Foliate
E: com.github.johnfactotum.foliate: python-script-but-no-python-dep usr/share/com.github.johnfactotum.Foliate/assets/KindleUnpack/compatibility_utils.py #!python
E: com.github.johnfactotum.foliate: python-script-but-no-python-dep usr/share/com.github.johnfactotum.Foliate/assets/KindleUnpack/mobi_split.py #!python
E: com.github.johnfactotum.foliate: python-script-but-no-python-dep usr/share/com.github.johnfactotum.Foliate/assets/KindleUnpack/mobiml2xhtml.py #!/usr/bin/python
E: com.github.johnfactotum.foliate: python-script-but-no-python-dep ... use --no-tag-display-limit to see all (or pipe to a file/program)
```

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -6,8 +6,8 @@ Build-Depends: debhelper, gettext, meson (>= 0.40), pkg-config, libglib2.0-dev (
 
 Package: com.github.johnfactotum.foliate
 Architecture: all
-Depends: gjs (>= 1.52), gir1.2-webkit2-4.0
-Recommends: python (>= 2.7) | python3 (>= 3.4), gir1.2-gspell-1
+Depends: gjs (>= 1.52), gir1.2-webkit2-4.0, python (>= 2.7) | python3 (>= 3.4) 
+Recommends: gir1.2-gspell-1
 Suggests: espeak, festival, dictd, dictd-dictionary, hyphen-af, hyphen-as, hyphen-bg, hyphen-bn, hyphen-ca, hyphen-cs, hyphen-da, hyphen-de, hyphen-el, hyphen-en-ca, hyphen-en-gb, hyphen-en-us, hyphen-es, hyphen-fi, hyphen-fr, hyphen-ga, hyphen-gl, hyphen-gu, hyphen-hi, hyphen-hr, hyphen-hu, hyphen-id, hyphen-is, hyphen-it, hyphen-kn, hyphen-lt, hyphen-lv, hyphen-ml, hyphen-mr, hyphen-nl, hyphen-no, hyphen-or, hyphen-pa, hyphen-pl, hyphen-pt-br, hyphen-pt-pt, hyphen-ro, hyphen-ru, hyphen-show, hyphen-sk, hyphen-sl, hyphen-sr, hyphen-sv, hyphen-ta, hyphen-te, hyphen-uk, hyphen-zu
 Description: Simple and modern eBook viewer.
  A simple and modern GTK eBook viewer, built with GJS and Epub.js.


apandada1:
Also, can we rename the debian package to just `foliate` instead of `com.github.johnfactotum.foliate`? We can change line 1 and 7 in the `debian/control` file to do it. We can also add a homepage so that people can go to http://https://github.com/johnfactotum/foliate/ if needed.

johnfactotum:
Thanks for the PR.

I don't think the Lintian error makes a lot of sense. Yes, we do have some Python scripts, but they are optional. They are only used for unpacking Kindle files. So it Recommends instead of Depends on Python. Even Lintian itself [mentions this](https://lintian.debian.org/tags/python-script-but-no-python-dep.html):

> In some cases a weaker relationship, such as Suggests or Recommends, will be more appropriate.

Seems to me that this is a Lintian bug? I suppose there's really no harm in making it Depends. But there's no real harm in leaving it as Recommends, either. Unless for some reason we really need to make Lintian happy?

As for the package name, originally I used the RDNN name for the deb because I planned on getting Foliate into elementaryOS's AppCenter. Of course that never got anywhere so it's not really needed at the moment. But there's really no advantage in changing it now either.

apandada1:
I was trying to make a ppa with automated build scripts (it is already working, and I was going to send another PR today after more testing). That would be just like the current den packages, but with automatic updates like snap, or the arch linux package. So it may be a good idea to minimize the lintian errors (as the python files are being shipped in the deb package). And sudo apt install foliate would be much easier to do.

apandada1:
This is the ppa. https://launchpad.net/~apandada1/+archive/ubuntu/foliate

To install Foliate from the ppa, use the following commands

```
https://launchpad.net/~apandada1/+archive/ubuntu/foliate
sudo apt update
sudo apt install com.github.johnfactotum.foliate
```

How it works- 
Launchpad automatically imports the code from https://github.com/johnfactotum/foliate.git, and it will build the package everyday if there are any new commits.
So, users can automatically update to the very latest code in GitHub.
Also, then it won't have to be manually packaged to .deb format.

I have tested it in two computers running KDE Neon and Linux Mint 19.3 (both with Ubuntu 18.04 base), but please test it, and then add it to the readme/website.

apandada1:
I can revert back python3 from dependency to `recommends`. But it will be helpful for the users if the debian package is renamed to just `foliate`. Shall I make another commit, making these changes?

johnfactotum:
Thanks a lot for your work on the PPA. Can you maybe apply the rename as patches in the build process?

Also, moving Python to Depends does not actually make the Lintian errors go way. I think it's more of a problem with the shebang. See https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=906211. But we don't even use the shebangs. Foliate manually uses either `python`, or `python3` if `python` isn't found, without regard to the shebang. So the Lintian error isn't really relevant.

Also note that the master branch is unstable and can break at any time. It's not meant to be used except for testing purposes. Most users will probably want to install the stable releases.

apandada1:
I have configured Launchpad to automatically do the build. I don't know how to rename it as a patch. Can we rename just the .deb package? (I don't know whether renaming `debian/control` will rename snap package etc.)
And, can you make another branch with the last stable release? I will configure Launchpad to use it.

apandada1:
The very latest code in the master branch seems to work well. So, for now, you can simply copy it to the a new `stable-releases` branch. Once the next stable release is ready, you'll have to merge the master branch into it.

johnfactotum:
I would rather not rename the package because (1) there's always the chance of confusion and breakage when renaming things, (2) if we ever want to publish on AppCenter, this is required, (3) not a lot of benefits from the name change, (4) distributions can always use any name they want; for example on MX Linux (a Debian based distro) they provide debs with the name `foliate`, and (5) since we're adding non-official packages, it's nice to have it namespaced so there's no chance of conflict.

> I don't know how to rename it as a patch. 

I don't know anything about PPAs. But it should be possible to add a simple patch to the source code. A lot of projects don't even have the `debian/` files upstreamed.

> And, can you make another branch with the last stable release?

Yeah, it's probably a good idea to have a separate branch for each minor release so we can release bug fixes. But it's kind of a hassle. Separate branch or not, it's probably best to point to specific commits/tags, rather than a whole branch. That is the usual way stable packages are maintained.

apandada1:
I think pointing to a commit can be done. Which one is the latest stable commit?
I will have to find how to apply the patch in Launchpad.

johnfactotum:
Every release is tagged. Please see https://github.com/johnfactotum/foliate/releases.

apandada1:
Done. The stable release is available at https://launchpad.net/~apandada1/+archive/ubuntu/foliate , and the daily builds are available at https://launchpad.net/~apandada1/+archive/ubuntu/foliate-daily. 
Please test it and add it to the readme/website. In the readme, it can be added against packages in Ubuntu (line 38, which right now points at the snap package while it points to native packages for other distributions. Snaps are already separately listed in the beginning of readme).

Once this is done, this pull can be closed. Lintian errors do not matter much anyway.

For now the name of the package is `com.github.johnfactotum.foliate`. I will see how to implement the build script to rename it in Launchpad.

johnfactotum:
I've added them to the readme. Thanks a lot for your work!

apandada1:
By the way, can I change the package name with a local commit in Launchpad?

johnfactotum:
> By the way, can I change the package name with a local commit in Launchpad?

Sure, if that's convenient for you. Maybe you can also add a note about the package name in the PPA's description, in case the readme here isn't updated.

> Also, since Manjaro and Arch Linux use almost the same package base, they can probably grouped in the readme.

Well, I think Foliate was actually added to the official Manjaro repo before it got into Arch, at least according to Repolgy. Anyway, the table is mostly sorted alphabetically.

PR #347, opened 2020-05-12
New zh_CN.po

bbblanccc:


bbblanccc:
and update zh_TW.po and es.po

po/zh_CN.po
```diff
@@ -0,0 +1,1948 @@
...
+
+#: src/window.js:183 src/window.js:184
+#, javascript-format
+msgid "of %d"
+msgstr "共 %d"
+
+#: src/export.js:31
+#, javascript-format
+msgid "<p>Annotations for</p><h1>%s</h1><h2>By %s</h2>"
+msgstr "<h1>《%s》</h1><p>的注释</p><h2>（%s 注）</h2>"
```
johnfactotum, line 79: This doesn't seem right. The second `%s` contains the author of the book, not the author of the annotations. Admittedly, the English text itself is ambiguous and confusing. But anyway that's just how it is now.

po/zh_CN.po
```diff
@@ -0,0 +1,1948 @@
...
+msgid "E-book Files"
+msgstr "电子书文件"
+
+#: src/main.js:90
+msgid "Open File"
+msgstr "打开文件"
+
+#: src/main.js:104#, fuzzy
+msgid "translator-credits"
+msgstr "翻译者致谢"
```
johnfactotum, line 56: The `translator-credits` string is used to credit translators in the "About" dialog. Should contain a list of translators' names separated by `\n` instead of the literal translation of "translator-credits".

po/zh_TW.po
```diff
@@ -37,7 +37,7 @@ msgstr ""
 
 #: data/com.github.johnfactotum.Foliate.desktop.in:12
 msgid "Ebook;Book;EPUB;Viewer;Reader;"
-msgstr "Ebook;Book;EPUB;Viewer;Reader;電子書;書;檢視器;閱讀器;"
+msgstr "電子書;圖書;EPUB;檢視器;閱讀器"
```
johnfactotum, line 40: This is the list of keywords of the application. Should probably keep the English keywords, because they are often useful to many users. See for example https://github.com/elementary/calculator/issues/61. It's also the common practice at least among GNOME apps.

Also, the list _must_ end with a semicolon.

bbblanccc: Thanks for pointing out, I will resovle these soon :smile:

PR #360, opened 2020-05-16
Update installation instructions in website

apandada1:
I have added the installation instructions in the website, and added Snap Store (which was missing in the installation instructions) as well. I also moved Flatpak and Snap to the right column to balance the length of the two columns
Here's how it looks now.
foliate

### DIFF
--- a/index.html
+++ b/index.html
@@ -207,17 +207,17 @@ footer a:link, footer a:visited {
         <div class="grid">
             <div>
                 <div>
-                    <h3>Flatpak</h3>
-                    <p><a href="https://flathub.org/apps/details/com.github.johnfactotum.Foliate"><img height="50" alt="Download on Flathub" src="flathub-badge-en.svg"></a></p>
                     <h3>Distribution packages</h3>
                     <p><a href="https://repology.org/project/foliate/versions"><img src="https://repology.org/badge/tiny-repos/foliate.svg" alt="Packaging status"></a></p>
                     <ul>
                         <li>Arch Linux: <a href="https://www.archlinux.org/packages/community/x86_64/foliate/"><code>foliate</code></a></li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -207,17 +207,17 @@ <h3>More Screenshots</h3>
...
                         <li>Arch Linux (AUR): <a href="https://aur.archlinux.org/packages/foliate-git/"><code>foliate-git</code></a></li>
                         <li>Fedora: <code>sudo dnf install foliate</code></li>
+                        <li>Ubuntu (Stable PPA): <a href="https://launchpad.net/~apandada1/+archive/ubuntu/foliate"><code>foliate</code></a></li>
+                        <li>Ubuntu (Daily builds): <a href="https://launchpad.net/~apandada1/+archive/ubuntu/foliate-daily"><code>com.github.johnfactotum.foliate</code></a></li>
                         <li>Void Linux: <code>xbps-install -S foliate</code></li>
                     </ul>
-                    <p>Debian/Ubuntu packages can be downloaded from the <a href="https://github.com/johnfactotum/foliate/releases">releases</a> page.</p>
+                    <p>Debian/Ubuntu packages can be downloaded from the <a href="https://github.com/johnfactotum/foliate/releases">releases</a> page. Note that it won't update itself. Use the <a href="https://launchpad.net/~apandada1/+archive/ubuntu/foliate">stable ppa</a> listed above to get automatic updates with a traditional package.</p>
```
johnfactotum, line 220: Thanks for the PR! I'm bit unsure about this bit, though. It kind of feels like giving too much endorsement to the PPA. The PPA works great, but it's neither maintained by the distribution, nor by the developer (me), so I would prefer not to give it this much weight.

apandada1: Ok, I will remove that line

apandada1: Done

PR #398, opened 2020-05-29
Fix GInotifyFileMonitor crash, fixes #357

### LINKED ISSUE #357: Crash with 'toggling down object GInotifyFileMonitor that's already queued to toggle up'
Foliate crashed on page turn. Happened for the first and only time of using it for more than a month.

The output of `journalctl /usr/bin/foliate` around the time the crash happened ([full log](https://pastebin.ubuntu.com/p/DG2nMT3gxS/)):

    May 15 16:58:15 enam-laptop com.github.john[115513]: toggling down object GInotifyFileMonitor that's already queued to toggle up
    May 15 16:58:15 enam-laptop audit[115513]: ANOM_ABEND auid=1000 uid=1000 gid=1000 ses=3 pid=115513 comm="com.github.john" exe="/usr/bin/gjs-console" sig=5 res=1

Distro: Arch Linux
`uname -a`: `Linux enam-laptop 5.6.11-arch1-1 #1 SMP PREEMPT Wed, 06 May 2020 17:32:37 +0000 x86_64 GNU/Linux`
Installed from: repository (version 2.1.1-1)

Rafostar:
Gio.FileMonitor has a bug that causes random GJS crashes when file is monitored for changes. The bug has been reported to the GJS devs but the cause is unknown and is still marked as unresolved. This commit fixes random foliate crashes by commenting out (disabling) file monitoring functionality until GJS bug is resolved.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -208,6 +208,7 @@ var Storage = GObject.registerClass({
         this.indent = indent
         this._file = Gio.File.new_for_path(path)
         this._data = this._read()
+        /* File Monitoring is currently broken: https://gitlab.gnome.org/GNOME/gjs/-/issues/297
         this._monitor = this._file.monitor(Gio.FileMonitorFlags.NONE, null)
         this._monitor.connect('changed', () => {
             if (this._getModified() > this._modified) {
@@ -216,6 +217,7 @@ var Storage = GObject.registerClass({
                 this.emit('externally-modified')
             }
         })
+        */
         this._debouncedWrite = debounce(this._write.bind(this), 1000)
     }
     static getPath(type, key, ext = '.json') {


Rafostar:
I think that temporary disabling file monitoring (with reloading files on changes) is still better than having random app crashes. The bug was originally reported here:
[https://gitlab.gnome.org/GNOME/gjs/-/issues/297](https://gitlab.gnome.org/GNOME/gjs/-/issues/297)

johnfactotum:
Thanks! Is there a way to reproduce the crash in Foliate? I think I've only ever seen the "toggling down" error once when importing a file, and it didn't crash, either.

Rafostar:
@johnfactotum 
I am not sure if enabling Gio file monitor to watch for file (and not doing any changes to that file) reproduces the problem on its own. But I can confirm that at least doing some changes to that file does. If you make a script that somehow changes the watched file from time to time (and causes the monitor to emit the 'changed' signal), then you should be able to reproduce it eventually. The bug is very random, sometimes it takes minutes to happen and sometimes hours!

johnfactotum:
Foliate does write to the monitored files when opening a book, and also when the reading location changes, so the files are being modified quite often. Occasionally it even crash as soon as the book is opened.

Still I would rather not turn file monitoring off. It breaks <kbd>Ctrl</kbd> <kbd>Shift</kbd> <kbd>R</kbd>, and (more importantly) "syncing" -- like, if a file synchronizing program has fetched remote changes, Foliate needs to be notified, otherwise the remote changes will be discarded as soon as you turn the page or resize the window, etc.

Obviously random crashing isn't good, either. But since the crash happens relatively infrequently, I would personally prefer having the file monitored.

Rafostar:
OK. I mostly wanted to help you identify the cause of GInotifyFileMonitor error. I know that working around bugs that aren't even yours is painful. You can close this PR if disabling file monitoring is a NO.

Maybe mentioning that this bug affects Foliate in GJS issue I linked would prompt the devs to pay more attention to it and fix it in near future.

johnfactotum:
Thanks!

I'll close this for now, then.

PR #410, opened 2020-05-30
Update zh_CN.po

bbblanccc:


po/zh_CN.po
```diff
@@ -1092,7 +1070,7 @@ msgstr "退出"
 
 #: src/ui/menuBar.ui:57
 msgid "_Edit"
-msgstr "编辑"
+msgstr "_编辑"
```
johnfactotum, line 1073: `_` is used to set the mnemonic key. For example, `_Edit` sets the key to "E", and the item will be activated when you press <kbd>Alt</kbd> + <kbd>E</kbd>. Because there's no key for "编" on the keyboard, this and others should be translated to something like `编辑(_E)`, which is the conventional way to translate mnemonic labels for CJK langauges.

bbblanccc: Sorry for late response & thanks for pointing out. It has been fixed.

PR #412, opened 2020-05-30
Fix Standard Ebooks OPDS link in catalog.js

acabal:
The URI points to the OPDS list of all ebooks, instead of pointing to the OPDS index. While the index currently only has one item ("all"), it may get new items in the future, like browse by tag or by most popular.


johnfactotum:
Good point. Thanks!

PR #490, opened 2020-06-24
Sidebar animation

### LINKED ISSUE #388: Add sidebar slide effect
Currently the sidebar mode just kind of "pops" in without any "slide" effect.

> I don't think this is currently possible with GTK. It is possible with `libdazzle`. I guess we can use `libdazzle`, but I'll want to make it optional, like the way we use `libhandy` at the moment; I don't want to hard-depend on a library just to add a few animations.

Yes, I think making it optional is a good idea. Do you know what [Setzer](https://github.com/cvfosammmm/Setzer) uses? _Peek_ didn't capture it very well but I think it shows the idea:

setzer_sidebarslide

johnfactotum:
Peek 2020-06-24 22-15

Fixes #388.


johnfactotum:
Closing for now, as it's clear that it'll not be merged. It will probably get animation once we switch to some libhandy widget for the sidebar.

PR #510, opened 2020-07-02
Updated Dutch translation

Vistaus:
:-)

### DIFF
--- a/po/nl.po
+++ b/po/nl.po
@@ -7,8 +7,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: com.github.johnfactotum.Foliate\n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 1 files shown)


johnfactotum:
Can you update the translations with the latest POT file?

Vistaus:
I already did, and the strings I now translated are the ones Poeditor showed as new. I updated the fork an hour ago, so the POT file is up-to-date.

johnfactotum:
But it's missing a lot of new strings. The POT creation date here is ` 2020-06-19 01:42+0800`. It should be `2020-07-02 14:22+0800`.

Vistaus:
Strange. Seems like GitHub had some issues (also noticed a few other oddities on the site). I now see a newer POT file. I'll translate the new strings now :)

Vistaus:
Alright, should be good to go now :)

johnfactotum:
Thanks!

PR #537, opened 2020-07-22
[Bug Fix] Do not use Wiktionary header as "language"

lpan:
# Context

As shown in the recording below, some of the internal links do not work.
foliate_bug_past_tense

# Investigation
I found that some of the internal links look like `https://en.wiktionary.org/wiki/tell#English`. As a result, the statement below will evaluate

```
internalLink.split('#') // => 'tell', 'English'
```

Since "English" is not a valid locale, `lookupFunc('tell', 'English')` will fail.

### DIFF
--- a/src/lookup.js
+++ b/src/lookup.js
@@ -128,10 +128,10 @@ const wiktionary = (word, language, lookupFunc) => lookup(
             if (internalLink && wikiNamespaces.every(namespace =>
                 !internalLink.startsWith(namespace + ':')
                 && !internalLink.startsWith(namespace + '_talk:'))) {
-                const [title, lang] = internalLink.split('#')
+                const [title] = internalLink.split('#')
                 const word = decodeURIComponent(title)
                     .replace(/_/g, ' ')
-                lookupFunc(word, lang || 'en')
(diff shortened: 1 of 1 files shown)


johnfactotum:
I've tried the words in your screenshot on [Quick Lookup](https://github.com/johnfactotum/quick-lookup), and it work fine, but in Foliate it doesn't, which is strange, because they pretty much use the same code: https://github.com/johnfactotum/quick-lookup/blob/651d4ae5a5eea19f37cd0dcda09bfac7cac5ab40/quick-lookup.js#L311-L314

So there's definitely a bug here, but you can't just assume that the language of the link is the same as the one that you're currently looking up. For example, when you look up a French word, it will often contain links to English entries.

johnfactotum:
OK, I figured it out. This is a regression caused by 72d5c701df3b3ad355581ef31e4a54c4af19cca1. The bug is caused by the language name "English" getting turned into "english" by `Intl.getCanonicalLocales` in the `getAlpha2` function.

johnfactotum:
Fixed in 2a4e81219f4fb6d23a85aaa2e4cf4412dbca6a33

PR #569, opened 2020-08-20
OPDS: Remember credentials

jhakonen:
Currently Foliate forgets OPDS credentials when the program exits, and prompts them again on next start, which is quite annoying, especially if you have a long password to type. This PR implements remembering of those credentials. 

Changes:
- Add "Remember my credentials" checkbox to the auth dialog. When checked it stores username and password to Secret Service. Checkbox is toggled by default if the credentials exist in Secret Service. Unchecking the checkbox lets user to use alternative credentials temporarily without affecting the stored credentials. To permanently remove stored credentials, user must remove them from the Secret Service using e.g. Seahorse.

I have tested this against my own COPS service I'm running on my NAS along along with Calibre's database, and it seems to work fine. Are there any corner cases that could break because of this that I should check?

Auth dialog:
auth-dialog

Deleting credentials from Seahorse:
seahorse

### DIFF
--- a/src/opds.js
+++ b/src/opds.js
@@ -554,7 +554,7 @@ var OpdsClient = class OpdsClient {
         const widget = this._widget
         const toplevel = widget instanceof Gtk.Widget
             ? widget.get_toplevel() : null
-        return promptAuthenticate(req, toplevel)
+        return promptAuthenticate(req, this.username, this.password, toplevel)
     }
     static opdsEntryToMetadata(entry, showSummary = true) {
         const {
--- a/src/opds.js
+++ b/src/opds.js
@@ -554,7 +554,7 @@ var OpdsClient = class OpdsClient {
         const widget = this._widget
         const toplevel = widget instanceof Gtk.Widget
             ? widget.get_toplevel() : null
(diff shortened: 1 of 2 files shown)


src/opds.js
```diff
@@ -554,7 +554,7 @@ var OpdsClient = class OpdsClient {
         const widget = this._widget
         const toplevel = widget instanceof Gtk.Widget
             ? widget.get_toplevel() : null
-        return promptAuthenticate(req, this.username, this.password, toplevel)
+        return promptAuthenticate(req, toplevel)
```
johnfactotum, line 557: I think the `this.username, this.password` bit is still needed for URLs like `http://username:password@example.com/`

johnfactotum (changes requested):

src/opds.js
jhakonen: Yea, looks like you're right. The auth dialog pops up unnecessarily with that kind of URL. Will fix.

jhakonen: Should be fixed now.

johnfactotum:
Thanks!

PR #591, opened 2020-09-08
Handle touchscreen events with Gtk.GestureSwipe

jhakonen:
Implements better touchscreen event handling. Swipe gesture detection is done with [Gtk.GestureSwipe](https://gjs-docs.gnome.org/gtk30~3.24.22/gtk.gestureswipe) instead of relying on webkit's wheel event. Moved wheel event handler from webkit page to GJS side, as there was no way to filter out touchscreen originating events within webkit. Implementation is based on discussion in #232

### DIFF
--- a/src/epubView.js
+++ b/src/epubView.js
@@ -641,6 +641,17 @@ var EpubView = GObject.registerClass({
                     } else if (velocityX < -SWIPE_SENSIVITY) {
                         this.goRight()
                     }
+                } else {
+                    // disable swipe up/down with non-paginated layouts (scrolled,
+                    // continuous), this protects against accidental switches if
+                    // user pans the page too fast
+                    if (!this.isPaginated) return
+
+                    if (velocityY > SWIPE_SENSIVITY) {
+                        this.prev()
+                    } else if (velocityY < -SWIPE_SENSIVITY) {
+                        this.next()
+                    }
                 }
             } catch (e) {
                 logError(e)
--- a/src/epubView.js
+++ b/src/epubView.js
@@ -22,7 +22,8 @@ const { EpubCFI } = imports.epubcfi
 const {
     debug, error, markupEscape, regexEscape,
     Storage, disconnectAllHandlers, base64ToPixbuf,
-    mimetypes, mimetypeIs, execCommand, recursivelyDeleteDir
+    mimetypes, mimetypeIs, execCommand, recursivelyDeleteDir,
+    setTimeout
 } = imports.utils
 
 const python = GLib.find_program_in_path('python') || GLib.find_program_in_path('python3')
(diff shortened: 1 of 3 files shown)


johnfactotum:
Thanks for the PR. I believe this is the right approach. And although I don't have a touchscreen to test with, enabling "Simulate touchscreen" in GTK Inspector actually does seem to work for the swipe gesture, and I can swipe by dragging.

Two minor issues, though:
1. It shouldn't be checking `this._webView.zoom_level`. Scrolling should still be enabled when the page is zoomed _by setting a zoom level_. Scrolling should, however, be disabled when the page is zoomed _by pinching_. The two "zoomings" are two completely different things. While WebKit provides pinch zooming out of the box, AFAIK it doesn't provide you with a property that you can check, which is what the old `getWindowIsZoomed` function was for.
2. With this PR, scrolling is now throttled, rather than debounced. Personally, I would much prefer the current behavior.

jhakonen:
1) Heh, I didn't even realize that there was support for pinch zooming, I just assumed that the zoom level was the only thing that was supported. At least I know now! And yes, my implementation was indeed wrong. I've changed the code so that the event handler now calls `getWindowIsZoomed()` function.

2) Changed as suggested. Originally, I tried to use the `debounce()`, but the `event` that the debounced function received had scroll deltas set to `0, 0`, That was the reason for `scrollX += deltaX`, etc... code, but I can just ignore events where the deltas are `0, 0`, it seems to work fine.

jhakonen:
I added check for `getWindowIsZoomed()` for the swipe gesture handler. Without it user might accidentally switch a page if the page is panned too fast. Swiping while pinch zoomed in `master` is disabled too, so now it should work same way as it does in `master`.

jhakonen:
What do you think, should the swipe gesture handler have a check for `isPaginated`?

Without it, it makes the scrolled layout easier to use with touchscreen, as tapping left/right side of the view does not switch pages. So you can use swipes instead.

With continuous layout, the swipes to left/right feels weird and a bit pointless as it just scrolls the web page up/down one screen height amount. In my opinion it feels just easier to pan (or swipe) the page up/down to scroll it.

jhakonen:
Added support for swiping up/down to switch a page. I had overlooked that you could do that in `master`.

BrainBlasted:
Tested locally. Seems to fix https://github.com/johnfactotum/foliate/issues/224 perfectly.

johnfactotum:
> What do you think, should the swipe gesture handler have a check for `isPaginated`?
> 
> Without it, it makes the scrolled layout easier to use with touchscreen, as tapping left/right side of the view does not switch pages. So you can use swipes instead.
> 
> With continuous layout, the swipes to left/right feels weird and a bit pointless as it just scrolls the web page up/down one screen height amount. In my opinion it feels just easier to pan (or swipe) the page up/down to scroll it.
> 
> In `master` scrolled and continuous layouts have the left/right swipes disabled. The behaviour would then be different in that regard with `isPaginated` left out from the gesture handler.

Yeah, I think swipes should definitely be disabled in continuous mode. Not really sure about the scrolled layout. It's quite convenient, and very intuitive, really. But I can also see that people might get annoyed if they intend to just scroll and accidentally makes it jump to a different chapter instead.

jhakonen:
Well, I could change it so that swiping doesn't work in continuous layout. That should at least cover that.

In scrolled mode the pages can be switched by touching the bottom panel's arrow buttons (using thumbs felt easiest to reach them), so not having the swipe support is not an issue. However, to change the page user still needs to move his hand to reach the buttons, which makes reading more cumbersome. And user would need both of his hands on the buttons in order to to switch pages back and worth, at least with tablet sized mobile device.

jhakonen:
Yay! 😄

Issue #590, opened 2020-09-07
Import/Export enhancement


itprojects:
It would be an enhancement, to add an option, to embed the JSON into the HTML as a BASE64 invisible string, thus the users, by default, would export/import from "HTML+JSON". This way people can have a very pretty HTML and a very functional JSON, each doing their own part.

@johnfactotum There could even be a pull request here...

itprojects:
Closing feature request; a simple HTML import may be more useful, and more fitting.

PR #593, opened 2020-09-09
Webkit HTML Import

itprojects:
HTML import feature. (Proposed new feature.)

~Export/Import book information to an HTML file, with the JSON information embedded in an invisible tag.~

~Aimed at issue https://github.com/johnfactotum/foliate/issues/590~

### DIFF
--- a/src/export.js
+++ b/src/export.js
@@ -13,9 +13,10 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
-const { Gtk, Gio } = imports.gi
+const { Gtk, Gio, WebKit2 } = imports.gi
+const ByteArray = imports.byteArray
 const ngettext = imports.gettext.ngettext
-const { mimetypes, readJSON, parseHTML, sepHeaderFunc } = imports.utils
+const { mimetypes, readJSON, sepHeaderFunc } = imports.utils
 const { EpubViewAnnotation } = imports.epubView
 const { EpubCFI } = imports.epubcfi
 const { AnnotationRow } = imports.contents
@@ -174,7 +175,7 @@ var importAnnotations = (window, epub) => {
     htmlFiles.set_name(_('HTML Files'))
     htmlFiles.add_mime_type('text/html')
 
-    const dialog = Gtk.FileChooserNative.new(
+    let dialog = Gtk.FileChooserNative.new(
         _('Import Annotations'),
         window,
         Gtk.FileChooserAction.OPEN,
@@ -184,78 +185,21 @@ var importAnnotations = (window, epub) => {
     dialog.add_filter(allFiles)
 
     if (dialog.run() === Gtk.ResponseType.ACCEPT) {
-        const file = dialog.get_file()
-
-        var json = null
         const isJSON = dialog.get_filename().endsWith(".json")
         const isHTML = dialog.get_filename().endsWith(".html")
 
+        const file = dialog.get_file()
+
+        window_global = window
+        epub_global = epub
+
         try {
             if (isJSON) {
-                json = readJSON(file)
+                let json = readJSON(file)
+                importWindowDialog(json)
             } else if (isHTML) {
-                json = parseHTML(file)
+                parseHTML(file)
             } else throw new Error()
-
-            if (!json.annotations.length) {
-                const msg = new Gtk.MessageDialog({
-                    text: _('No annotations'),
-                    secondary_text: _('There’s nothing to import.'),
-                    message_type: Gtk.MessageType.INFO,
-                    buttons: [Gtk.ButtonsType.OK],
-                    modal: true,
-                    transient_for: window
-                })
-                msg.run()
-                msg.destroy()
-                return
-            }
-            const list = new Gio.ListStore()
-            let annotations = json.annotations
-                .sort((a, b) => EpubCFI.compare(a.value, b.value))
-                .map(({ value, color, text, note }) =>
-                    new EpubViewAnnotation({
-                        cfi: value,
-                        color: color || 'yellow',
-                        text: text || '',
-                        note: note || ''
-                    }))
-
-            annotations.forEach(annotation => list.append(annotation))
-
-            const builder = Gtk.Builder.new_from_resource(
-                '/com/github/johnfactotum/Foliate/ui/importWindow.ui')
-            const dialog = builder.get_object('importDialog')
-            dialog.transient_for = window
-
-            const [width, height] = window.get_size()
-            dialog.default_width = Math.min(500, width)
-            dialog.default_height = height * 0.7
-
-            const listbox = builder.get_object('annotationsListBox')
-            listbox.bind_model(list,
-                annotation => new AnnotationRow(annotation, epub, false))
-            listbox.set_header_func(sepHeaderFunc)
-
-            if (dialog.run() === Gtk.ResponseType.ACCEPT) {
-                if (!json.metadata || !json.metadata.identifier
-                || json.metadata.identifier !== epub.metadata.identifier) {
-                    const msg = new Gtk.MessageDialog({
-                        text: _('Identifier mismatch'),
-                        secondary_text: _('The identifier of the imported file does not match the book’s. This might indicate that the annotations do not belong to this book.'),
-                        message_type: Gtk.MessageType.QUESTION,
-                        modal: true,
-                        transient_for: dialog
-                    })
-                    msg.add_button(_('Cancel'), Gtk.ResponseType.CANCEL)
-                    msg.add_button(_('Import Anyway'), Gtk.ResponseType.ACCEPT)
-                    const accept = msg.run() === Gtk.ResponseType.ACCEPT
-                    msg.destroy()
-                    if (!accept) annotations = null
-                }
-            }
-            dialog.close()
-            return annotations
         } catch (e) {
             const msg = new Gtk.MessageDialog({
                 text: _('Error'),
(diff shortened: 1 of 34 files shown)


johnfactotum:
Using `` seems wrong semantically (it's not really an image).

And it's really not that difficult to parse the HTML as is; I'm not sure how adding JSON to it would help anything. Maybe should just add code to parse the HTML instead.

itprojects:
@johnfactotum The  tag can be changed to anything you like. What would you prefer?

Yes, just adding code to parse HTML would be more natural, and backwards compatible. However, that takes time to process, it slows the app down. The JSON variant is a copy/paste matter; can  safely be done "on UI thread".

The idea to use Base64 was introduced to protect the JSON contents from syntax clashes, anywhere along the process.

Which direction is more desireable to you? JSON option, or HTML parsing?

itprojects:
@johnfactotum Switched to an invisible `<p>` with two patches. Is this better?

johnfactotum:
> However, that takes time to process, it slows the app down. The JSON variant is a copy/paste matter; can safely be done "on UI thread".

I kind of doubt that. Even if it is slower to parse the HTML, it'd hardly be the bottleneck when importing (which I guess would be rendering all the `GtkListBox`es for the annotations, and also adding the highlights in Epub.js).

In Firefox, for example, when you choose "Backup", it exports your bookmarks to a JSON file, but they also allow you to import/export with only HTML.

> Which direction is more desireable to you? JSON option, or HTML parsing?

The ability to import from HTML would be nice. I don't think there's any downside to that. If nothing else it can act as a fallback when no JSON data is not embedded. I also feel that "HTML (+JSON)" can be potentially confusing to the user.

> Switched to an invisible `<p>` with two patches. Is this better?

I think the proper way is to use a `<script>` tag, with `type="application/json"`. This is similar to how JSON-LD is embedded.

itprojects:
@johnfactotum  If you don't want the "HTML (+JSON)", we can add it as a tick box - "Also embed JSON in HTML".



It would be great to see an HTML parser!, but the serializer has problems:

1. A JSON format has identifiers, so checks can be added that we're working on the right book, before data is damaged.

2. Serializer does not specify class for notes or highlighted text. Example: `cfi`, `section`, but no `quote` or `note`.

If this is accepted as a bug, and fixed, then some notes can be imported, but some older-standard notes could not be.

3. Serializer makes line endings into `<br>`, which looks great, but can potentially cause information loss.

The replacement is a good idea, as it just looks better.

See: exportToHTML,`note.replace(/\n/g, '<br>')`.

In JSON:

`"note":"Note1.\n<br>\nNote2."`

In Foliate:

Note1.
`<br>`
Note2.

In HTML:

`Note1.<br><br><br>Note2.`

Most books won't be affected, but some computer technology books can be.

One ugly fix is to add an `id` global tag, `<br id="html-tag"/>`, which looks wrong, and will create a new standard.

src/utils.js
```diff
@@ -78,9 +78,18 @@ var regexEscape = str => str ? str.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&') : ''
 var readJSON = file => {
```
johnfactotum, line 78: I don't think one should modify the `readJSON()` function. This function is for reading all JSON files, and JSON files only. If you want to read embedded JSON data when importing, then you should change the import function.

src/utils.js
```diff
@@ -78,9 +78,18 @@ var regexEscape = str => str ? str.replace(/[.*+\-?^${}()|[\]\\]/g, '\\$&') : ''
...
     try {
         const [success, data, /*tag*/] = file.load_contents(null)
-        if (success) return JSON.parse(data instanceof Uint8Array
-            ? ByteArray.toString(data) : data.toString())
-        else throw new Error()
+        if (success) {
+            var raw = data instanceof Uint8Array ? ByteArray.toString(data) : data.toString()
+            if (raw.includes('json-db')) {// quick check
+                try {
+                    var found = raw.match(/<script id="json-db" type="application\/json">(.*)<\/script>/)[1]
```
johnfactotum, line 85: I don't really like how it looks for `/<script id="json-db" type="application\/json">(.*)<\/script>/`. I would probably prefer properly parsing the HTML. But, even if one doesn't want to do that, one still needs to first distinguish between JSON and HTML files, otherwise it would fail if someone were to include `<script id="json-db" type="application/json"></script>` in their notes.

src/export.js
```diff
@@ -14,13 +14,18 @@
...
 const ngettext = imports.gettext.ngettext
 const { mimetypes, readJSON, sepHeaderFunc } = imports.utils
 const { EpubViewAnnotation } = imports.epubView
 const { EpubCFI } = imports.epubcfi
 const { AnnotationRow } = imports.contents
 
-const exportToHTML = async ({ annotations }, metadata, getSection) => {
+const exportToHTML = async (jsonData, metadata, getSection, withJson) => {
+    const json_base64 = withJson ? 
+        `<script id="json-db" type="application/json">${GLib.base64_encode(JSON.stringify(jsonData))}</script>` : ''
```
johnfactotum, line 26: You don't have to base64 encode the JSON if you put it in `<script>`; `JSON.stringify()` should be enough.

johnfactotum (changes requested):
>  If you don't want the "HTML (+JSON)", we can add it as a tick box - "Also embed JSON in HTML".

Yes, I think that would be much better.

> A JSON format has identifiers, so checks can be added that we're working on the right book, before data is damaged.

This isn't really a big problem; identifiers aren't always reliable, anyway. And one can also add checks for title and author. A more reliable check -- the only reliable check, actually -- is to get the text for the CFI and match it with the quoted text.

It's true that the JSON file includes more information than the HTML file, but those information are currently irrelevant as far as importing is concerned, as Foliate can only import annotations at the moment, not bookmarks, reading progress, etc.

> Serializer does not specify class for notes or highlighted text. Example: cfi, section, but no quote or note. If this is accepted as a bug, and fixed, then some notes can be imported, but some older-standard notes could not be.

The quote is enclosed in the `<blockquote>` tag, plus there's styling for highlight color, and the paragraph that follows, which doesn't have any attributes, is the note. It's entirely unambiguous and easily parsable.

> In JSON: `"note":"Note1.\n<br>\nNote2."` In Foliate: `Note1. <br> Note2.` In HTML: `Note1.<br><br><br>Note2.`

This is simply a bug. The HTML exporter is currently broken; if you add JSON to it, it is _still_ broken.

> One ugly fix is to add an id global tag, `<br id="html-tag"/>`, which looks wrong, and will create a new standard.

It _is_ wrong.
- First of all, you can't use `id`, because an `id` must be unique within a document.
- Secondly, what if someone decides to write `<br id="html-tag">` in their notes? Then it would still be broken.
- Thirdly, and most importantly, it makes no sense to convert `\n` to `<br>` _and_ at the same time allow HTML tags in notes. It's one or the other. It should either stop converting `\n` to `<br>` altogether, _or_ simply escape `<` and `>` when exporting to HTML.

src/utils.js
itprojects: Agreed

src/utils.js
itprojects: Yes, that's true.

src/export.js
itprojects: The idea was to avoid the `<br>` breaking the code. This practice seems to be standard.

itprojects:
@johnfactotum The work to make an HTML parser has now started...

src/export.js
johnfactotum: `<br>` won't break anything. `</script>` will. So one should just escape `</script>`, rather than base64 encoding.

Besides, if base64 is used, one should probably do something like
```
<script type="application/json">
{
   data: "${base64 string here}"
}
</script>
```

Because a naked Base64 string is not valid JSON.

itprojects: The Base64 gives some data integrity.
How would you escape `</script>` ?

itprojects:
@johnfactotum The HTML parser now has a working file reader and parser.

Here's some problems you can advise about:

`<h1>Book Title</h1><h2>By Author A. Author</h2>`

the word `By ` is an issue, because it conflicts checking the author's name,

see in French:
`<p>Annotations pour</p><h1>%s</h1><h2>Par %s</h2>`,

but `Par ` is 4 letters and the check will return false.

One way to change this would be:
`<h1>Book Title</h1>By<h2> Author A. Author</h2>`
where the starting space can be stripped.

itprojects:
@johnfactotum An initial HTML Parser code has been comitted to this pull request. A review would be helpful.

itprojects (commented):
Those changes are relevant to the JSON feature.

itprojects:
@johnfactotum Bump. Changes made...

johnfactotum:
Can't say that I like parsing HTML with regex very much... Preferably it should be parsed properly, like with a browser, since we already use WebKit, anyway.

itprojects:
@johnfactotum Can you suggest gists of code from Foliate? Any code addition (even directly) is more than welcome!

itprojects:
@johnfactotum After testing WebKit HTML parsing code, it seems that the **code flow is very async**, and **very slow** because of the WebView, and requires **more than twice the lines**. On the other hand using a **regex is linear, quick, and debuggable**.

How do you want to proceed?

WebKit or Regex?

johnfactotum:
> it seems that the code flow is very async

That is presumably a good thing, not a bad thing.

> and very slow because of the WebView,

Admittedly I have not tested it, but is it really that much slower? The bottleneck when importing is when actually adding the annotations, rather than parsing.

> and requires more than twice the lines.

That's an exaggeration, surely? Maybe a bit more lines, but nothing too terrible.

> On the other hand using a regex is linear, quick, and debuggable.

Yeah, but regex is not reliable. Ideally it should use a simple HTML parser, rather than a full-blown browser. But AFAIK there's not much in way of that in GJS currently.

src/utils.js
```diff
@@ -86,6 +86,54 @@ var readJSON = file => {
...
+    .replace(/[\n]/g, '\\n')
+    .replace(/[\r]/g, '\\r')
+    .replace(/[\t]/g, '\\t');
+};
+
+var parseHTML = file => {
+    try {
+        const [success, data, /*tag*/] = file.load_contents(null)
+        if (success) {
+            var raw = data instanceof Uint8Array
```
johnfactotum, line 106: Please follow the code style and don't use `var` unless you're exporting the variable. Use `const` if possible, or `let` if you need to mutate it.

src/utils.js
```diff
@@ -86,6 +86,54 @@ var readJSON = file => {
...
+                    note_text = note_text.replace(/<br>/g, '\n')
+                } catch (e) {}
+
+                if (note_cache.length != 0)
+                    note_cache += ','
+
+                note_cache += `{"value":"${escapeJSON(note[1])}","color":"${note[2]}","text":"${escapeJSON(note[3])}","note":"${note_text == null ? '' : escapeJSON(note_text)}"}`
+            }
+
+            return JSON.parse("{\"annotations\":[%s]}".format(note_cache))
```
johnfactotum, line 128: Why are you building a JSON string and parsing it? Just build an object.

Also please don't use `String.prototype.format()` unless the string needs to be translated.

src/utils.js
```diff
@@ -86,6 +86,54 @@ var readJSON = file => {
...
+    .replace(/[\"]/g, '\\\"')
+    .replace(/[\/]/g, '\\/')
+    .replace(/[\b]/g, '\\b')
+    .replace(/[\f]/g, '\\f')
+    .replace(/[\n]/g, '\\n')
+    .replace(/[\r]/g, '\\r')
+    .replace(/[\t]/g, '\\t');
+};
+
+var parseHTML = file => {
```
johnfactotum, line 102: IMO this function should live in `export.js` because it's not a generic HTML parsing function. It's specific to parsing exported HTML.

itprojects:
@johnfactotum Please, please - a simple unambiguous yes or no. In one word, should this Pull Request be updated for RegEx (as per review)?

If there is a bottleneck with annotation adding, with WebKit there will be a new (second) bottleneck; based on testing experience with small samples (5 annotations). Are you willing to wait 30 min for only 80 notes (typical usage)?

Is it really that much more code? **YES**. In fact the (WebKit) code is not even entirely complete; it can extract, but to pass them to Foliate will need even more to come.

Currently the WebKit variant is not online, but it is so slow and async (because of WebKit), that there will be a necessity to modify export.js functions.

Is RegEx reliable? _No_. But if reliable is your choice (without WebKit) there is still the option to put the annotations in a Base64 tag, and ignore both RegEx and WebKit, since then we only need JSON.parse on a safe string. Only down side is that this make backward compatibility only after re-export to HTML from a wisely saved JSON.

johnfactotum:
> Please, please - a simple unambiguous yes or no. In one word, should this Pull Request be updated for RegEx (as per review)?

I have already stated my preference against the regex solution above. The review I just gave mostly concerns general coding issues and not really related to regex per se.

> Are you willing to wait 30 min for only 80 notes (typical usage)? [...] Is it really that much more code? YES. In fact the (WebKit) code is not even entirely complete; it can extract, but to pass them to Foliate will need even more to come.

It's hard to tell without seeing and running the code, but I think you're definitely doing something wrong if you're getting that kind of performance.

itprojects:
@johnfactotum That's good, so we do it with WebKit. A new branch (for WebKit version) will be setup to make it easier to see where things may have gone wrong. Your input is much appreciated, as always.

src/utils.js
itprojects: Yes, indeed.

src/utils.js
itprojects: That is a very good point. The function was put in utils.js mostly because there was some JSON code there.

src/utils.js
itprojects: Good to know.

itprojects:
@johnfactotum Code for the HTML import was improved, and is now expected to work for both HTML+JSON imports without major problems. Check this branch: `importhtmlwebkit` . Test it out; merge ready?

itprojects:
@johnfactotum Is this still relevant?

itprojects:
@johnfactotum This Pull Request has the completed feature of Import HTML notes (with webkit). However, _it seems_, that for some reason is not part of the Foliate development plan. Unless you wish to merge it, it will be closed.

johnfactotum:
Yeah, sorry. There are a lot things that I'd like to see changed before this can be merged. But I don't really have the time or energy for the whole review process right now.

Also, while experimenting with this a while ago I realized that the whole import code is a bit necessarily complicated and unreliable, so I'd like to work on that instead before adding HTML import.

So I guess I will close this for now. Still, thanks for your contribution!

PR #602, opened 2020-09-30
Fix offlint dictionaries lookup

chaitu-tk:
As the initialization of dictionaries is declared const, further
dictionaries are not being added, removing const to add any offline
dictionaries.

### DIFF
--- a/src/lookup.js
+++ b/src/lookup.js
@@ -308,7 +308,7 @@ const parseStarDictDbs = d => d.split('\n').filter(d => d).map(row => {
     return { name: cols[0] }
 })
 
-const dictionaries = {
+dictionaries = {
     wiktionary: {
         name: _('Wiktionary (English)'),
         noWrap: false,


johnfactotum:
`const` does not make the object immutable. Is offline dictionaries not working for you? Because it works for me.

chaitu-tk:
> const does not make the object immutable.
Yes, the objects can change just we cannot reassign. But for some reason, I was only able to see "Wiktionary". 

> Is offline dictionaries not working for you?
Yes, with this fix I am able to see the offline dicts in the "lookup dictionary"

chaitu-tk:
I guess it's probably because of the scope of "const" those additions are ignored.

johnfactotum:
Well, like I said, offline dictionaries are working for me. And I don't see how this change would affect anything. Can you maybe provide a bit more information? Like what system, versions you're using, and what kind of offline dictionaries are you using, and so on?

chaitu-tk:
Sure, I am using `dictd` as an offline dictionary on Ubuntu 18.04 system and I have built foliate from sources (`bb7cd52`), reproed the issue, and with the fix issue got resolved.

I just tried to reproduce the issue using `master`, but not seeing it anymore, not sure why? Is there a persistent state/cache I can remove to repro it?

As an aside, I can see the below JS warning in my system in both cases.

`js-Message: 15:34:26.035: JS WARNING: [resource:///com/github/johnfactotum/Foliate/js/lookup.js 315]: assignment to undeclared variable dictionaries`

johnfactotum:
> I just tried to reproduce the issue using master, but not seeing it anymore, not sure why?

Did you restart Foliate the first time you added offline dictionaries? Offline dictionaries are only added on startup, so if you have just installed an offline dictionary, you need to restart Foliate to use it. Perhaps that's the reason?

chaitu-tk:
> Did you restart Foliate the first time you added offline dictionaries? Offline dictionaries are only added on startup, so if you have just installed an offline dictionary, you need to restart Foliate to use it. Perhaps that's the reason?

I guess I did, anyway, as the code looks correct closing this PR now. Thanks, for the inputs.

PR #706, opened 2021-04-04
Update cs.po

pafri:
Update of translation (mostly the strings with reports about fixes)

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -1,21 +1,21 @@
 # Copyright (C) YEAR THE PACKAGE'S COPYRIGHT HOLDER
 # This file is distributed under the same license as the com.github.johnfactotum.Foliate package.
 #
-# Pavel Fric <pavelfric@seznam.cz>, 2019, 2020.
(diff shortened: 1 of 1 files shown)


pafri:
**Note**: in Mageia distribution, there is some issue with the **path**, where the .mo file is stored. I don't know, if it originates from packagers work, or comes from the Foliate locale settings? The **observation**: after launch of Foliate the application's GUI remains in English. Also, I took a look, where the .mo file connected with Foliate localization is located and found, that it is in this path:

**/usr/share/locale/cs_CS/LC_MESSAGES/**

Here is placed the file "com.github.johnfactotum.Foliate.mo", but doesn't affect the way the interface looks, as mentioned above.

So, I've created .mo file with that naming (com.github.johnfactotum.Foliate.mo) and placed it in the path below (where I would expect that file originally; maybe named like "foliate.mo"):

**/usr/share/locale/cs/**

Now the Foliate's GUI is/texts are localized.

johnfactotum:
It used to use `cs_CS` but this has been fixed with 01e98b38161e347cdbe27c3c8a2074896a65d803, which is included in 2.6.x.

Issue #631, opened 2020-11-20
Map the "Esc" keyboard shortcut to "Go back to the library view"


nekohayo:
When inside a book, I'd like to be able to easily go back to the library/list of books without needing to munch through a hamburger menu.

I think the Escape key would be the standard/expected keyboard shortcut to use for this.

JohannPopper:
Hmm. Yes. And/or a discoverable library button on the header. I agree accessing the library should be a quick function.

itprojects:
There is a discussion for this. https://github.com/johnfactotum/foliate/discussions/656

johnfactotum:
Fixed with 42388fd4bd1d1c98d0c8aa65b278d5df300c106d

PR #776, opened 2021-09-05
Add shortcut to get back to the library

guille:
Fix #656

I used Ctrl+H since Escape (suggested in #631 was already in use to get out of Fullscreen. I'm using this locally and it seems to be working, but first contrib so let me know if I'm missing something.


johnfactotum:
Thanks!

I'm still not sure about the shortcut. I was thinking maybe backspace could work, though I'm not sure if that'd break anything. But I guess I'll merge this, and people can still give feedback on the specific shortcut if they want.

PR #802, opened 2021-10-16
Stop unescaping entities on pango markup

### LINKED ISSUE #799: Reference preview not showing due to unescaped entities
**Describe the bug**
I noticed the reference preview feature was not working in some of my epubs (the popup shows but the text is empty). I finally got around to investigating and found this error in the logs:

```
(com.github.johnfactotum.Foliate:5544): Gtk-WARNING **: 19:20:54.280: Failed to set text '<sup>[2]</sup> footnote text<a href="Text/Cap2.xhtml#rf2"><<</a>' from markup due to error parsing markup: Error on line 1 char 201: “<<” is not a valid name
```

The footnote is using `<<` as the text for the link to go back to the text.

I suspect this is related to #495 (as well as the fix for that), but I'm missing context on how to fix this properly. Maybe you have some ideas @johnfactotum. Just changing `src/web/utils.js` as so

```diff
-    return unescapeHTML(doc.body.innerHTML.trim()).replace(/&/g, '&amp;')
+    return unescapeHTML(doc.body.innerHTML.trim()).replace(/&/g, '&amp;').replace(/</g, '&lt;')
```

Would fix the error but I imagine it would reintroduce #495.

**To Reproduce**
1. Modify any epub file with footnotes, add `<<` to the text.
2. Try to see the reference preview for it

**Version:**
 - Foliate version: 2.6.3

guille:
Fixes #799

After the changes, footnotes with "<" in them show up fine in the preview:

I'd appreciate a review @johnfactotum since this is partially a revert of fc8f972

### DIFF
--- a/src/web/utils.js
+++ b/src/web/utils.js
@@ -118,5 +118,5 @@ const toPangoMarkup = (html, baseURL = '') => {
         })
         if (nodeName === 'a' && !el.hasAttribute('href')) usurp(el)
     })
-    return unescapeHTML(doc.body.innerHTML.trim()).replace(/&/g, '&amp;')
+    return unescapeHTML(doc.body.innerHTML.replace(/&lt;/g, '&amp;lt;').trim()).replace(/&/g, '&amp;')
 }


guille:
@johnfactotum it's been a while, sorry. I kept thinking I'd have time to look into the "proper solution" but I never did, so I've just pushed a commit with the workaround you suggested above.

johnfactotum:
Thanks!

johnfactotum:
Now after b557b0d82b99572a8743c6583de3b8d29ca706e1, footnotes are finally rendered properly, with all the contents and stylesheets intact, exactly as they would be rendered in the main view.

This proper solution is indeed quite involved. It's made possible with the [new renderer](https://github.com/johnfactotum/foliate/issues/962). Unlike Epub.js, the new renderer decouples loading and rendering. Thus one can easily create additional views from an already loaded book.

guille:
That's great, @johnfactotum . Thanks a lot for your continued efforts on Foliate.

PR #860, opened 2022-01-24
Set ToC direction from book direction

### LINKED ISSUE #859: Make ToC respect the book's language direction
**Is your feature request related to a problem? Please describe.**
For the time being, Table of Contents (ToC) is displayed as LTR regardless of the book's language direction. When viewing a book with RTL lang. dir. (which implies that its ToC text is most likely RTL too since the high likelihood of the ToC being the same language as the book), the experience was a bit off since the ToC is not displayed properly as it is supposed to be—the ToC treeview (and its scrollwindow parent) should be right aligned instead of left aligned.

**Describe the solution you'd like**
Set ToC direction based on the book's language direction.

**Describe alternatives you've considered**
–

**Additional context**
Actually, such implementation has already been carried out on Foliate's navbar (145d7bfce1c9882ec2e187f36fcf7cee74a7ad8a) and footer (5ca3d12075d9974a5ac7e13f29fefff8687b3091).

fathyar:
Closes #859

### DIFF
--- a/src/contents.js
+++ b/src/contents.js
@@ -120,6 +120,14 @@ var ContentsStack = GObject.registerClass({
     }
     set epub(epub) {
         this._epub = epub
+
+        this._epub.connect('metadata', () => {
+            const rtl = this._epub.metadata.direction === 'rtl'
+            const dir = rtl ? Gtk.TextDirection.RTL : Gtk.TextDirection.LTR
+            this._tocTreeView.get_parent().set_direction(dir)
+            this._tocTreeView.set_direction(dir)
+        })
+
         this._tocTreeView.model = this._epub.toc
 
         this._updateData(epub.annotations, epub.bookmarks)


johnfactotum:
Thanks for the PR!

I'm not unsure whether this is the best way to determine direction for the TOC. The `direction` in Epub.js is [actually](https://github.com/futurepress/epub.js/blob/f6fece4f88a8f7a8b03a0bf3863fc8b75ff21166/src/packaging.js#L62) just [`page-progression-direction`](https://www.w3.org/publishing/epub32/epub-packages.html#attrdef-spine-page-progression-direction), which may or may not be the correct direction for the TOC.

For example, I think there are fixed layout comic books that have RTL page progression, but horizontal LTR text, in which case the TOC should be LTR. Another problem is vertical writing for CJK languages. The page progression will be RTL if the writing mode is `vertical-rl`, but the TOC is always rendered horizontal (ideally it should be vertical be that's a whole different issue) and should (usually) be LTR. This is actually also an issue for the footer, though it's less obvious there as the footers are aligned to the center.

I think the most correct way is probably to check the direction of the [navigation document](https://www.w3.org/publishing/epub32/epub-packages.html#sec-package-nav) (and possibly even the direction of each item!). According to its [type definition](http://www.daisy.org/z3986/2005/ncx-2005-1.dtd), the NCX file in EPUB 2 also supports the `dir` attribute (though I'm not sure how many books actually specifies this). And if no direction is specified there, one would need to make a guess based on the language, which, fortunately, is very easy to do using [`Intl.Locale.prototype.textInfo`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Locale/textInfo). Unfortunately one has to patch Epub.js to get the dir and lang attributes.

PR #888, opened 2022-04-16
Fix "zoom" translation to spanish.

KJ2ME:
The word "zum" is an "anglisismo" and can be improved changing it with another better translations.

Almost the time it is rare to read "zum", it is even better to read it in English and of course, have better ways to say it without use an "anglisismo".

Sorry for use "anglisismo", idk how to say it on english.


johnfactotum:
Thanks!

> Sorry for use "anglisismo", idk how to say it on english.

I think it's "anglicism".

PR #896, opened 2022-05-05
Update fr translations

mbousq:
Couple of fixes and translations.
As a note: perhaps "Cache locations" could be renamed "Cache progression" ?

```
#: src/ui/preferenceWindow.ui:700
#: data/com.github.johnfactotum.Foliate.gschema.xml:28
msgid "Cache locations"
```


johnfactotum:
>  As a note: perhaps "Cache locations" could be renamed "Cache progression" ?

Yes, that sounds like a good idea. But might not be worth changing at this point, as the string has been stable for so long, and when I port to GTK 4, I plan on ditching Epub.js and calculate progression based on the raw filesize data of the sections, which would eliminate the need for caching.

PR #899, opened 2022-05-17
Adding Hebrew translation

yarons:
Most of the untranslated strings are either very hard to translate to Hebrew or require gender bias.


johnfactotum:
Thanks!

PR #913, opened 2022-06-24
Update Standard Ebooks OPDS URL

acabal:
We've recently changed the URL of our OPDS feeds. The old URL redirects with HTTP 301, but we may as well update the source in catalog.js with the new URL to avoid an extra redirect.


johnfactotum:
Thanks!

PR #936, opened 2022-09-05
Add StartupNotify to desktop file

Newbytee:
GTK apps have built-in support for this, and it allows shells like
Phosh to show a splash screen while Foliate is starting.

Reference: https://developer-old.gnome.org/integration-guide/stable/startup-notification.html.en


johnfactotum:
Thanks!

PR #1027, opened 2023-05-05
Update architectures part in snap yaml to allow build for armhf,arm64 and amd64 for snap packages

CapeCrusader321:
@johnfactotum  with this it will try to build for all architectures and will build for arm-hf and 64 and amd64 while will fail for two other architectures, i am still finding a way to limit builds to armhf, arm64 and amd64


CapeCrusader321:
@johnfactotum  This is fixed, merge it when free ....

PR #1034, opened 2023-05-30
updated to latest commit of gtk4 branch

soumyaDghosh:
This is also using webkitgtk-6, but it's in beta and supports only amd64 for now.

### DIFF
--- a/snapcraft.yaml
+++ b/snapcraft.yaml
@@ -17,7 +17,7 @@ layout:
     symlink: $SNAP/gnome-platform/usr/bin/gjs
 
 slots:
-  dbus-daemon:
+  foliate:
     interface: dbus
     bus: session
     name: com.github.johnfactotum.Foliate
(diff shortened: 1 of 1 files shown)


johnfactotum:
The binary name is now `foliate`. Also there's no TTS support, so espeak, etc. is unused at the moment. The plan was to talk to host speech-dispatcher through socket. Alternatively it might use WebKitGTK's speech synthesis API. At any rate it doesn't and probably won't run espeak directly.

soumyaDghosh:
Does foliate uses portals?

johnfactotum:
If you mean when opening files, yes. GTK uses portals automatically.

soumyaDghosh:
Then, no use of `home` and `removable-media`

soumyaDghosh:
Any thoughts on merging this?

johnfactotum:
Well, if you're sure that v2.6.4 stable works and won't need to be updated anymore, I guess. Or is there a way to maintain both versions?

soumyaDghosh:
One single snap can do the trick, v2.6.4 is only need to updated if you make changes there.

soumyaDghosh:
Seems like, it doesn't builds, when done from snap-store. Fixing it. Sorry about that typo.

Issue #1036, opened 2023-05-30
Question about the gtk4 branch


soumyaDghosh:
**Question:**

As the snap is not able to provide proper gjs path, so, we need to manually do it. Now, in this [file](https://github.com/johnfactotum/foliate/blob/gtk4/src/main.js), there are three places, that take `@GJS@`, which parts needs to be replaced with `/usr/bin/gjs`?

**Version:**
 - Foliate version: latest, git
 - OS/Distribution and version: Ubuntu 22.04
 - Desktop environment: GNOME 44
 - Installation method: Snap

johnfactotum:
There's a comment that says "the latter would be replace by Meson". Meson ignores it when `@` is preceded by `\`.

soumyaDghosh:
So, only the first line needs the change? AFAIU

soumyaDghosh:
webkitgtk needs one fix, a bug, I have found. Thanks to your app, I got it.

soumyaDghosh:
![image](https://github.com/johnfactotum/foliate/assets/72045785/988cb26a-2a9a-4aee-97c2-dae7c47f863e)

Should I push this?

soumyaDghosh:
You will need to have a forum request, under store request category, to autoconnect with webkitgtk. Else, users will have to manually connect to it using:

`sudo snap connect foliate:webkitgtk webkitgtk`

johnfactotum:
If you could make the request as a contributor/maintainer of the package, that would be great. To be honest it seems bizarre to me that this is even necessary.

soumyaDghosh:
Yup.. cause webkitgtk content snap is made by me, not by someone from canonical or core snap team. Though, every snap requires store request to get the auto connect granted. BTW, can you add as a collaborator in that package. It'd be easy for me then. 

My email id: soumyadeepghosh2004@zohomail.in

PR #1037, opened 2023-05-31
Fixed snap for gtk4 branch with webkitgtk-6.0

soumyaDghosh:
As said in #1036, working perfectly, as the WebKitGtk is still in beta, so, the app should also be in beta.


soumyaDghosh:
Screenshot from 2023-05-31 01-08-55

Issue #1048, opened 2023-06-13
Timestamp in annotation


azzamsa:
**Is your feature request related to a problem? Please describe.**
Yes. I always wanted to know my reading progress. How my thought developed over the years. One of them is to see the timestamp in highlighted text or a note in the pdf/epub files.

**Describe the solution you'd like**
The annotations should contain a timestamp.
Either right-click on the annotation list on the sidebar or hover over the mouse.

**Describe alternatives you've considered**
- 

**Additional context**
Okular was able to show the timestamp in PDF annotation. But the ebook has to be saved in `.okular` archive format, which can't be managed via a book manager such as Calibre. Because it is unlikely for them to support the `.zip` format. `.okular` is just a `.zip`.

![image](https://github.com/johnfactotum/foliate/assets/17734314/ffcb427d-ed8e-48e2-b582-6683861aff81)

I have checked the `.json` file of Foliate, and currently, it doesn't save any information about the date and time.

![image](https://github.com/johnfactotum/foliate/assets/17734314/e57667b5-62ce-48d9-a818-415037bda642)

```json
  "lastLocation": "epubcfi(/6/8!/4/2[sbo-rt-content]/2/18/1:0)",
  "progress": [
    14,
    657
  ],
  "annotations": [
    {
      "value": "epubcfi(/6/10!/4/2[sbo-rt-content]/2/4,/5:2,/5:39)",
      "color": "lime",
      "text": "Jon has been teaching Rust since 2018",
      "note": ""
	 "timestamp": // ⚠️ <-- please support timestamp
    }
```

PR #1049, opened 2023-06-14
Add timestamp to annotation

azzamsa:
Hi.

It is a working implementation, but still very rough. I am ready to take any guidance to polish this PR further.

image

Fixes: #1048

### DIFF
--- a/src/annotations.js
+++ b/src/annotations.js
@@ -17,6 +17,8 @@ const Annotation = utils.makeDataClass('FoliateAnnotation', {
     'color': 'string',
     'text': 'string',
     'note': 'string',
+    'created': 'string',
+    'modified': 'string',
 })
 
 const AnnotationHeading = utils.makeDataClass('FoliateAnnotationHeading', {
@@ -462,16 +464,21 @@ export const AnnotationPopover = GObject.registerClass({
             actions: ['add-note', 'delete', 'more'],
         }))
 
+        const currentDateTime = new Date().toISOString();
+        this.annotation.set_property('created', currentDateTime)
+
         this._drop_down.selectColor(this.annotation.color)
         this._text_view.buffer.text = this.annotation.note
         this.#updateStack()
 
         this._drop_down.connect('color-changed', (_, color) => {
             this.annotation.set_property('color', color)
+            this.annotation.set_property('modified', currentDateTime)
             this.emit('color-changed', color)
         })
         this._text_view.buffer.connect('changed', buffer => {
             this.#updateStack()
+            this.annotation.set_property('modified', currentDateTime)
             this.annotation.set_property('note', buffer.text)
         })
     }


johnfactotum:
- Ideally it should probably have both `created` and `modified` instead of a single timestamp.
- The date should be saved as an ISO string.
- The displayed date should use the user's locale setting. See `src/format.js`.
- Opening the annotation popover should not change the date. A new date should be created when modifying the annotation.
- It might be better to display the dates in the popover instead of the sidebar.

azzamsa:
I would like to separate this into two commits. 1) Storing to the JSON file 2) display in pop-over. I would like to ask your opinion on the first one. If it's okay, then I will start working on the second step.

azzamsa:
It is very hard for me do debug because the right click is not reliable.

```
(com.github.johnfactotum.Foliate:11213): Gtk-CRITICAL **: 23:02:55.082: _gtk_css_corner_value_get_y: assertion 'corner->class == &GTK_CSS_VALUE_CORNER' failed

(com.github.johnfactotum.Foliate:12531): Gtk-WARNING **: 23:06:20.941: gtk_widget_size_allocate(): attempt to allocate FoliateBookItem box 0x55a
9d5bd50e0 with width -30 and height -18

(WebKitWebProcess:2): GStreamer-WARNING **: 16:06:30.247: External plugin loader failed. This most likely means that the plugin loader helper bi
nary was not found or could not be run. You might need to set the GST_PLUGIN_SCANNER environment variable if your setup is unusual. This should
normally not be required though.
```

Every time I add a new annotation the app freeze. I need to restart every time I want to debug.

```
OS: Fedora Linux 38 (Container Image) x86_64
Kernel: 6.3.8-200.fc38.x86_64
Shell: zsh 5.9
DE: GNOME
WM: Mutter
```

```
⬢ ❯ dnf list installed | grep gtk4
gtk4.x86_64                                 4.10.4-1.fc38                      @updates
javascriptcoregtk4.0.x86_64                 2.40.2-1.fc38                      @updates
javascriptcoregtk4.0-devel.x86_64           2.40.2-1.fc38                      @updates
webkit2gtk4.0.x86_64                        2.40.2-1.fc38                      @updates
webkit2gtk4.0-devel.x86_64                  2.40.2-1.fc38                      @updates

⬢ ❯ dnf list installed | grep webkit
webkit2gtk4.0.x86_64                        2.40.2-1.fc38                      @updates
webkit2gtk4.0-devel.x86_64                  2.40.2-1.fc38                      @updates
webkitgtk6.0.x86_64                         2.40.2-1.fc38                      @updates

⬢ ❯ dnf list installed | grep libadwaita
libadwaita.x86_64                           1.3.3-1.fc38                       @updates

⬢ ❯ gjs --version
gjs 1.76.2
```

johnfactotum:
A new popover is created each time you click on an existing annotation. So the created date should not be set in the popover's constructor.  It should be set when creating the annotation. And new timestamps should be created when modifying the annotation.

> It is very hard for me do debug because the right click is not reliable.

There's no right click action anywhere in the app, though. Well, except for built-in context menu for GtkEntry and GtkTextView.

> Every time I add a new annotation the app freeze.

Can't reproduce this. It works fine for me.

johnfactotum:
So I've went ahead and added this myself. Still thinking about how to present it in the UI.

PR #1050, opened 2023-06-14
Update README

azzamsa:
In Fedora-based distros. All the package mentioned is correct, but the webkitgtk.

In Fedora, it is `webkitgtk6-0`. Or should I mention both Fedora and Debian dependencies?

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Read books in style.
 - `gjs` (>= 1.72)
 - `gtk4` (>= 4.10)
 - `libadwaita` (`gir1.2-adw-1` in Debian-based distros)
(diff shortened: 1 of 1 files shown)


johnfactotum:
It's still called `webkitgtk-6.0` in Arch Linux and in WebKit's [documentation](https://github.com/WebKit/WebKit/blob/main/Source/WebKit/gtk/migrating-to-webkitgtk-6.0.md). I *think* the dash was present at some point in Fedora, too. At least there used to be a `webkit2gtk-4.0` package. They probably changed it to distinguish the API version more clearly from the actual version of the package (e.g. `2.40.1`), though I can't find anything on this change.

azzamsa:
Oh, I see.

Should we close this PR then?

PR #1051, opened 2023-06-14
Add missing `mesa runtime` to README

azzamsa:


azzamsa:
When I want to make a PR. I installed `gjs, gtk4, libadwaita, webkitgtk-6.0`, but I get an error saying that I have no mesa runtime. Installing `mesa-libEGL-devel` solves the issue.

johnfactotum:
Sounds like a packaging bug of WebKitGTK. Foliate doesn't use Mesa directly so it shouldn't be listed as a dependency.

PR #1065, opened 2023-07-30
 Restore GitHub Action for building Debian package

johnfactotum:


.github/workflows/deb.yml
```diff
@@ -0,0 +1,34 @@
...
+  build:
+    # The type of runner that the job will run on
+    runs-on: ubuntu-latest
+
+    # Steps represent a sequence of tasks that will be executed as part of the job
+    steps:
+      # Checks-out your repository under $GITHUB_WORKSPACE, so your job can access it
+      - uses: actions/checkout@v2
+
+      - run: sudo apt install build-essential debhelper meson gettext pkg-config libglib2.0-dev gjs appstream
```
archisman-panigrahi, line 28: In my computer, it does not build without the new build dependencies added [here](https://github.com/johnfactotum/foliate/pull/1123) (they were not necessary in Foliate version 2).

archisman-panigrahi: Ah, now it does not build because Ubuntu 22.04 repositories have too old versions of gjs (and probably other software too). We probably need to wait until 24.04 GitHub actions for it to generate a deb package.

johnfactotum: Or add an option to skip the dependency checks. But that makes no sense, either. Those dependencies are supposed to be run-time dependencies. Either we're using Meson wrong and should specify run-time dependencies another way, or Meson should be skipping these checks automatically when building a package.

johnfactotum: I don't think we're doing anything wrong, cause Meson says:
```
Run-time dependency gjs-1.0 found: YES 1.78.0
Run-time dependency gtk4 found: YES 4.12.3
Run-time dependency libadwaita-1 found: YES 1.4.0
Run-time dependency webkitgtk-6.0 found: YES 2.42.2
```

So they are *run-time* dependencies indeed.

This means it could be the same bug that requires you to install `desktop-file-utils`.

jbicha:
I believe I can help here.

Why do you want to build a Debian package? There are different reasons and they might mean we handle this in different ways. Here are some:
- Just checking if Foliate is buildable on Debian
- Distribute nightly builds
- Make it easier to install on your computer

Also, the commit message says Debian but this is actually being built on Ubuntu.

johnfactotum:
The action existed in v2.x. It was mainly for providing a deb package (nightly or otherwise), though it did also function as a check.

> Also, the commit message says Debian but this is actually being built on Ubuntu.

Well, I used the term "Debian package" to refer to the deb format. But it was mainly intended for Ubuntu.

archisman-panigrahi:
@johnfactotum It should work with the new 24.04 based builder https://github.com/github/roadmap/issues/958

archisman-panigrahi:
The action seems to be stuck for some reason https://github.com/johnfactotum/foliate/actions/runs/8918176680/job/24492363460?pr=1065

jbicha:
Adding Ubuntu 24.04 CI is on Github's roadmap. I haven't seen an announcement that it is enabled yet.

archisman-panigrahi:
The GitHub action is working after 1. installing build dependencies, 2. fixing debhelper-compat version, and 3. checking out the submodules. Here is my code. https://github.com/archisman-panigrahi/foliate/pull/1 
@johnfactotum Should I send a PR?

johnfactotum:
@archisman-panigrahi Yes, please. Thanks! Also I see there's a deprecation notice on actions/upload-artifact v2. Is that fixable by simply changing the version to v4?

archisman-panigrahi:
> Is that fixable by simply changing the version to v4?

Yes it got fixed this way. 

Here is my PR https://github.com/johnfactotum/foliate/pull/1383

PR #1083, opened 2023-09-26
Added Orgmode to annotations export menu

MateusSartorio:
Added the option to export annotations in Org Mode format.


src/annotations.js
johnfactotum, line 1: `toast` was changed to `add_toast` in 541a487b58f7485fb1bd2aa331411d633ba1fb50.

johnfactotum:
Thanks!

PR #1097, opened 2023-10-14
Fixes for foliate snap

soumyaDghosh:


snapcraft.yaml
```diff
@@ -55,4 +58,10 @@ parts:
       - usr/share/metainfo/com.github.johnfactotum.Foliate.metainfo.xml
     organize:
       snap/foliate/current/usr: usr
+  deps:
+    plugin: nil
+    stage-packages:
+      - libspeechd2 # probably necessary, need testing
+    prime:
+      - usr/lib/*/libspeechd.so.*
```
johnfactotum, line 66: I don't think this is needed. It doesn't use libspeechd.

soumyaDghosh: Firefox snap also has it. And as mentioned, needs much more testing.

johnfactotum: Well, Firefox is probably using libspeechd, i.e. speech-dispatcher's official C client library. Foliate can't and doesn't use that. It has its own client implementation in GJS.

PR #1116, opened 2023-11-13
UI: Mark a few strings translatable

ghost:
Add missing translatable strings.

Also the '100%' should be marked as translatable for some languages.


johnfactotum:
Thanks! The "100%" should actually be deleted from the UI file and formatted with `Intl` instead.

PR #1123, opened 2023-11-13
added Debian build dependencies

archisman-panigrahi:
I added the necessary build dependencies so that it builds, and also uploaded packages to the Launchpad PPA (for Ubuntu 23.10 and 24.04), which are building right now.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,7 +7,10 @@ Build-Depends: debhelper-compat(=10),
 	meson (>= 0.59),
 	pkg-config,
 	libglib2.0-dev (>= 2.54),
-	gjs,
+	libgjs-dev,
+	libgtk-4-dev,
+	libadwaita-1-dev,
+	libwebkitgtk-6.0-dev,
 	appstream
 Standards-Version: 4.5.0
 Rules-Requires-Root: no
--- a/debian/control
+++ b/debian/control
@@ -11,6 +11,7 @@ Build-Depends: debhelper-compat(=10),
 	libgtk-4-dev,
 	libadwaita-1-dev,
 	libwebkitgtk-6.0-dev,
+	desktop-file-utils,
 	appstream
 Standards-Version: 4.5.0
 Rules-Requires-Root: no


johnfactotum:
TBH I don't understand why these are necessary.

Is `desktop-file-utils` just for running the post install `update_desktop_database` thing? But isn't Meson supposed to skip that when building debs?

archisman-panigrahi:
Without `desktop-file-utils`, the Launchpad build failed with the meson error 
```
ERROR: Program 'update-desktop-database' not found or not executable
```
From https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=1022550 I found that adding `desktop-file-utils` fixes this issue.

Also, meson fails to compile the debian package without the `-dev` packages for gtk4, libadwaita and gjs, and libwebkitgtk-6.0.

johnfactotum:
> Also, meson fails to compile the debian package without the `-dev` packages for gtk4, libadwaita and gjs, and libwebkitgtk-6.0.

Ah yes, I think that's probably due to the dependency checks in 6148752efeda58a759a30214c7eb5b73841fd8ea?

> Without `desktop-file-utils`, the Launchpad build failed with the meson error

Yeah, and that looks like a Meson and/or dpkg bug to me. IIUC, the post install script is supposed to run only when installing, not when building a package.

archisman-panigrahi:
> Ah yes, I think that's probably due to the dependency checks in [6148752](https://github.com/johnfactotum/foliate/commit/6148752efeda58a759a30214c7eb5b73841fd8ea)?

Yes, I think so.

> Yeah, and that looks like a Meson and/or dpkg bug to me. IIUC, the post install script is supposed to run only when installing, not when building a package.

I don't know much about this to have an opinion, but could be a bug.

johnfactotum:
From https://mesonbuild.com/Gnome-module.html:

>  If `DESTDIR` is specified during installation all scripts will be skipped.

So probably for whatever reason, `DESTDIR` is no longer set when building debs. It surely worked before for Foliate v2.0, which has a post install script that checks for `DESTDIR` as well: https://github.com/johnfactotum/foliate/blob/master/build-aux/meson/postinstall.py. And it worked; you didn't have to install `desktop-file-utils`.

Issue #269, opened 2020-04-06
"Invert" color theme distorts sub-pixel font rendering


jspatz:
The "invert" option is very handy for dealing with ebooks that include hard-coded colors.  However, it appears to be inverting text colors _after_ sub-pixel font rendering is applied, causing the sub-pixel colors themselves to be inverted, and therefore creating significant color-fringing distortions.

Example:

Text with "light" theme":
![image](https://user-images.githubusercontent.com/12176759/78605548-c27a2b00-7829-11ea-9595-48bac9bd2118.png)

Same text with "invert" theme:
![image](https://user-images.githubusercontent.com/12176759/78605619-eb022500-7829-11ea-910b-c6b1afe46d01.png)

For comparison, "dark" theme, which does not exhibit this issue:
![image](https://user-images.githubusercontent.com/12176759/78605666-ff462200-7829-11ea-941e-0b6f5f3ace4c.png)

johnfactotum:
I don't think there's much we can do, as the rendering is done using CSS filters. And unfortunately, that just seems to be how filters work. For example, if you pinch zoom after applying the filter, the text gets blurry. There's probably a WebKit bug about this somewhere, or maybe that's just how filters are supposed to work.

For the moment I'll mark this as not our bug. For other methods of overriding book styles, we'll track it in https://github.com/johnfactotum/foliate/issues/267.

johnfactotum:
Actually I think this can be avoided by inverting only the text but not the background. This can be demonstrated with the following example:
```html
<div style="background: #fff; color: #000; filter: invert(1) hue-rotate(180deg);">
    <h1>Test</h1>
</div>
<div style="background: #000;">
    <h1 style="color: #000; filter: invert(1) hue-rotate(180deg);">Test</h1>
</div>
```

![image](https://github.com/johnfactotum/foliate/assets/50942278/3a5f1ce9-330f-459a-a868-3867c5283ebc)

PR #1137, opened 2023-11-16
Disable subpixel smoothing when the reader is going to be inverted.

colons:
The 'Invert' theme inverting text after subpixel smoothing has already been applied was causing some pretty unpleasant-to-look-at chromatic aberration.

The `gtk4` branch does not appear to have this theme. It seems like maybe it's intended to come back, based on the styles in `reader.html`, but I'm not going to try to fix this there since I can't test it.

Before:
dark-prod

In this branch:
dark-antialiased

### DIFF
--- a/src/web/epub-viewer.js
+++ b/src/web/epub-viewer.js
@@ -302,6 +302,7 @@ const setStyle = style => {
             'background': bgColor,
             'font-size': `${fontSize}px !important`,
             'line-height': `${spacing} !important`,
+            '-webkit-font-smoothing': invert ? 'antialiased' : 'auto',
             '-webkit-hyphens': hyphenate ? 'auto' : 'manual',
             '-webkit-hyphenate-limit-before': 3,
             '-webkit-hyphenate-limit-after': 2,


johnfactotum:
Thanks! Didn't know about this property. One question: does setting it to `antialiased` force antialiasing for users that disable it in fontconfig or gnome-tweaks? (Or does WebKitGTK not respect those to begin with?)

> The `gtk4` branch does not appear to have this theme.

It does, but it's now a setting called "Invert Colors in Dark Mode" that you can enable no matter which theme you're using.

The GTK 4 version also does inversion a bit differently in that only the book's contents are inverted. So the background (which is now the parent page behind the iframe that's displaying the book) is not inverted. See #269.

colons:
> One question: does setting it to antialiased force antialiasing for users that disable it in fontconfig or gnome-tweaks? (Or does WebKitGTK not respect those to begin with?)

I don't know, and I'm not entirely sure how I'd check. For what it's worth, I nothing else I use has subpixel smoothing enabled; one of my monitors is sideways and another is a 4k monitor running at 1080p; it wouldn't make sense in my setup. But Foliate has it enabled in the reader view anyway.

> It does, but it's now a setting called "Invert Colors in Dark Mode" that you can enable no matter which theme you're using.

Oh, okay, I didn't spot that. I'll try to make another MR doing this there.

colons:
Ah, okay, I'm able to see this effect in some parts of books in the gtk4 branch, although I'm unable to make it go away by putting `-webkit-font-smoothing: antialiased;` in the reader styles in places I would have expected it to work. Is there a way to get to webkit's web inspector within foliate so I can see what's going on?

dark-warning-gtk4

This is less of a big deal for me in the gtk4 branch, though, for this book at least, since these warning boxes don't look as broken in the default dark view.

johnfactotum:
You can open the inspector with F12 or from the view menu (the three dots menu on the headerbar above the page).

johnfactotum:
Fixed in cfcfa6d2c00230f9e8a35483451cffb08ec7c1f6. Seems to working for me.

PR #1244, opened 2024-02-02
Reduced size of screenshot.png so that README loads more quickly

Sparrowkeet:
Please reject this PR if not helpful! My internet is slow, so I noticed that screenshot.png loaded slowly in the README. It's a compelling image, and the updated screenshot_smaller.png does not significantly reduce the quality (in my opinion). I used tips from this article to reduce the filesize: https://logosbynick.com/gimp-reduce-file-size-png/


johnfactotum:
The shadows are pretty important. Perhaps it would be better to switch to WebP, which, with the `-lossless` argument would yield a smaller size compared to the optimized PNG while retaining the shadows.

But this sort of optimization seems a bit unnecessary for the README. Probably worth doing for the website, though, which has an even bigger screenshot.

Sparrowkeet:
Thank you so much for the feedback! If swapping out the website PNG for a WebP file would be helpful, I'm happy to do that

johnfactotum:
Thanks. I suppose, then, the screenshots for the website should be converted to webp. But should probably leave the screenshots in the readme until Flathub and GNOME Software, etc. support webp (not sure if they do now) because those are the same ones included in the appdata, and duplicating them just for the readme doesn't seem to be worth it to me.

Sparrowkeet:
See my comment on the commit for info on the file conversion process. Thanks for the opportunity to help, and no worries if you end up not using these for any reason!

johnfactotum:
Sorry, I meant that we should keep the PNGs in the README and only convert the images in the `gh-pages` branch.

The screenshots should probably be updated, too, rather than merely converted, because right now it shows the old version with the button labeled "Library", which has since been replaced by an icon button. But that's a separate issue.

johnfactotum:
Updated the website in c8f5e6788989e66b01d7e8ef5589280bb6ed8e7f. I leaving the screenshots in the README until there's confirmation that WebP is supported by GNOME Software, etc.

PR #1335, opened 2024-06-07
Add Spiel speech API

eeejay:
[Spiel](https://project-spiel.org/) is a modern speech synthesis API for the desktop that will hopefully support many kinds of providers and voices. It has GI bindings, so adding it to foliate shouldn't be hard.

I started a port, but ran into some trouble with how foliate creates SSML mark elements to report speech progress. Spiel has speech boundary events, including SSML marks, but now all providers support it. Unfortunately I think changes will be needed in foliate-js as well to make this work. Specifically, pre-segmenting the text into marks gets in the way here.

### DIFF
--- a/src/gresource.xml
+++ b/src/gresource.xml
@@ -13,6 +13,7 @@
     <file>search.js</file>
     <file>selection-tools.js</file>
     <file>speech.js</file>
+    <file>spiel.js</file>
     <file>toc.js</file>
     <file>themes.js</file>
     <file>tts.js</file>
--- a/src/spiel.js
+++ b/src/spiel.js
@@ -1,0 +1,47 @@
+import Gio from 'gi://Gio'
+import Spiel from 'gi://Spiel'
+
+Gio._promisify(Spiel.Speaker, 'new')
+
+export class SpielClient {
+    #speaker
+    #pitch = 1.0
+    #rate = 1.0
+    #promises = new Map()
+    async init() {
+        if (!this.#speaker) {
+          this.#speaker = await Spiel.Speaker.new(null);
+        }
+    }
+
+    speak(str) {
+        const text = str.replace('\r\n.', '\r\n..') + '\r\n.'
+        console.log(`speak: "${text}"`);
+        const rate = this.#rate;
+        const pitch = this.#pitch;
+        const utterance = new Spiel.Utterance({text, rate, pitch, is_ssml: true});
+        let speaker = this.#speaker;
+        return new Promise((resolve, reject) => {
+          speaker.connect('utterance-finished', (_, utt) => { if (utt == utterance) resolve(); });
+          speaker.connect('utterance-canceled', (_, utt) => { if (utt == utterance) resolve(); });
+          speaker.connect('utterance-error', (_, utt, err) => { if (utt == utterance) reject(err); });
+          speaker.speak(utterance);
+        });
+    }
+    pause() {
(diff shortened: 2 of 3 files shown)


johnfactotum:
If you mean supporting providers that don't support mark events but do support word boundary events, that is indeed not something currently supported by foliate-js. I do plan on adding this, since SSML support seems to be problematic in browsers. (Ideally I'd like to switch to the Web Speech API once it's supported by WebKitGTK.)

One slight snag is that the marks are currently also used to implement "speak from here" and pausing, mainly to ensure that the speech always begins from word boundaries. Maybe it could do without this (or decouple this from the speech text), in which case it shouldn't be too hard to do without marks. Just need to maintain a text walker instance (see https://github.com/johnfactotum/foliate-js/blob/main/text-walker.js) to convert string offsets to ranges.

eeejay:
It makes sense moving into the web view. Happy to see that it is at least [proposed in WebKitGTK](https://bugs.webkit.org/show_bug.cgi?id=270742).

Note: because of the way spiel works there is no need to track the speech progress for pausing purposes. Calling pause will simply pause the stream. So I think SSML marks would need to stay exclusively for speech dispatcher support. And yeah, we would need to de-serialize the string offset to a DOM range, which may or may not be trivial?

johnfactotum:
> Calling pause will simply pause the stream.

Ideally I think it should be capable of rewinding to the start of the last word or sentence other than simply pausing the stream. The current behavior in Foliate isn't good either because it simply restarts from a word boundary, which would result in incorrect pronunciation or intonation. The same goes for "speak from here" (i.e. starting speech from a user-selected position).

Avoiding part words isn't really an important feature, though. Mostly it's just what you get "for free" since the text is already segmented.  It's probably fine to just drop it.

> And yeah, we would need to de-serialize the string offset to a DOM range, which may or may not be trivial?

For plain text it should be okay. It might be non-trivial when using SSML, when the offsets are that of the SSML source string, because in general mapping source offsets to nodes is difficult if not impossible with the browser's DOM APIs. But since the XML here is controlled and rather simple, maybe one can just count the number of characters between `<` and `>` and adjust the offsets accordingly.

johnfactotum:
I noticed that the upstream issue has been fixed: https://github.com/WebKit/WebKit/commit/290d009f7b35d6af9bf5b5b80d113f5db6a04274. So I guess going forward, one should fix this by implementing all TTS functionality in foliate-js with the Web Speech API. The implementation should ideally have the exact same interface as how foliate-js currently handles media overlays. That would simplify things a lot.

ColorfulCode728:
> [Spiel](https://project-spiel.org/) is a modern speech synthesis API for the desktop that will hopefully support many kinds of providers and voices. It has GI bindings, so adding it to foliate shouldn't be hard.
> 
> I started a port, but ran into some trouble with how foliate creates SSML mark elements to report speech progress. Spiel has speech boundary events, including SSML marks, but now all providers support it. Unfortunately I think changes will be needed in foliate-js as well to make this work. Specifically, pre-segmenting the text into marks gets in the way here.

It says `(com.github.johnfactotum.Foliate:265224): Gjs-CRITICAL **: 22:36:54.532: JS ERROR: Error: Requiring Spiel, version none: Typelib file for namespace 'Spiel' (any version) not found` when running foliate. But i did install spiel from flatpak repo. I used this pr to build.

PR #1336, opened 2024-06-10
Add space/shift+space shortcuts for turning pages

### LINKED ISSUE #1283: Add Space/Shift+Space shortcuts for turning pages
I miss being able to use Space to turn pages (and Shift+Space to turn backwards). The space bar is larger than the arrow keys and the central location makes it easier to to press.

Most (all?) other ebook reader apps I've tried support these shortcuts, such as Apple Books and Google Play Books. I believe the pre-GTK4 version of Foliate also supported this shortcut.

Note that I certainly don't want to *replace* any existing shortcuts, just add these.

elcste:
Resolves https://github.com/johnfactotum/foliate/issues/1283

I tested this change with a local snapcraft build.


johnfactotum:
They should be mapped to `view.prev` and `view.next` so that it would work with RTL books.

elcste:
Embarrassing as a one-time linguistics student!

I made that change and tested with a RTL book (https://github.com/IDPF/epub3-samples/releases/download/20230704/israelsailing.epub).

PR #1378, opened 2024-08-11
Added polish translation

Repcake:
Added polish translation


Repcake:
Hopefully everything is correct, its my first time contributing to anything on github. The app is amazing and the only issue i had was no polish translation.

PR #1381, opened 2024-08-27
Hide an inspector button from the navbar

korpe1:
Why do we need to reveal an inspector button for regular users?
image

And an stupid question, what the earth this "Web inspector" thing even do in the Foliate? I'm just curious :)


johnfactotum:
EPUB and Kindle formats are based on Web technologies. So it's the same as web browsers exposing an inspect button in the menu.

Issue #1333, opened 2024-06-03
Is there any way to configure TTS in Foliate 3.1.1?


hjings:
**Question:**
I want to use edge-tts with Foliate. But I couldn't find a way to configure the TTS in Preferences (like the screenshot in [this link](https://github.com/johnfactotum/foliate/issues/618)). I could only find the Narration button on the nav bar, which provides limited functions. Is there any way to configure?

**Version:**
 - Foliate version: 3.1.1
 - OS/Distribution and version: Linux Mint 21.3 Cinnamon
 - Desktop environment: Cinnamon 6.0.4
 - Installation method: Flatpak

johnfactotum:
In v3 you cannot change the TTS command in Foliate. It functions as a Speech Dispatcher client. To change the TTS engine, you need to configure Speech Dispatcher.

archisman-panigrahi:
@hjings Yes you can. Please see https://askubuntu.com/q/1526184/124466

PR #1391, opened 2024-09-09
Add how to change TTS voice to FAQ

archisman-panigrahi:


archisman-panigrahi:
This should fix #1333

PR #1399, opened 2024-09-15
Fix Gtk.Popover select translate and save state select language.

keygenqt:
my_gif

Added saving of language selection for translation, it is very inconvenient to select a new translation language each time. In addition to this, there is a bug related to Gtk.Popover & select (after selecting a language, the window can only be closed with focus on select, I have not yet figured out how to overcome this, it looks like a gtk bug).

If it is possible to save the state in the application without using localStorage, I will be glad if you tell me.

### DIFF
--- a/src/selection-tools.js
+++ b/src/selection-tools.js
@@ -249,16 +249,43 @@ html, body {
     width: 100%;
 }
 body {
+    flex: 1;
     display: flex;
     flex-direction: column;
 }
-select {
+header {
+    flex: 1;
+    padding: 8px;
+    display: none;
+}
+header button {
     width: 100%;
+    margin-top: 6px;
+}
+header button:first-child {
+    margin-top: 0;
 }
 main {
     flex: 1;
+    display: flex;
     overflow-y: auto;
     margin: 8px 0;
+}
+footer {
+    display: flex;
+    justify-content: space-between;
+    flex-direction: row;
+    width: 100%;
+}
+footer div {
+    white-space: nowrap;
+}
+footer a,
+footer a:link,
+footer a:visited,
+footer a:hover,
+footer a:active {
+    color: black;
 }
 #output {
     padding: 0 8px;
@@ -266,7 +293,10 @@ main {
 </style>
 <header></header>
 <main><div id="output"></div></main>
-<footer>${_('Translation by Google Translate')}</footer>
+<footer>
+    <div>${_('Translation by Google Translate')}</div>
+    <a id="language" href="javascript:void(0)">?</a>
+</footer>
 <script>
 const googleTranslate = (text, lang = defaultLang) => {
     fetch('https://translate.googleapis.com/translate_a/single?client=gtx'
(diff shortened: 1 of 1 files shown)


marsaif:
thank you so much for adding this feature

johnfactotum:
Using localStorage is okay (in theory using GSettings would be better but probably not worth the trouble).

But I would like to prevent it from leaking into other contexts, if possible (preferably each tool should have a separate origin). That's mainly why I haven't gotten around to fixing this. I *think* I attempted to set a different origin by setting the URL to `foliate:selection-tool`... but apparently that doesn't work.

But maybe it doesn't matter. But perhaps it should at least use a different URI scheme, to separate it from the main `foliate:` scheme.

keygenqt:
my_gif


Now that's ok. I don't really understand GJS, as I understand the problem Gtk.Popover + Gtk.DropDown/select is not new https://gitlab.gnome.org/GNOME/gtk/-/issues/5568. I fixed Popover by replacing select with buttons and add saving via localStorage. Maybe I didn't do everything perfectly, but it works =)

src/selection-tools.js
```diff
@@ -249,24 +249,54 @@ html, body {
...
+}
+footer div {
+    white-space: nowrap;
+}
+footer a,
+footer a:link,
+footer a:visited,
+footer a:hover,
+footer a:active {
+    color: black;
```
johnfactotum: Don't hardcode the color, which would break in dark mode. Use something like `initial`, `inherit`, or `currentcolor`.

Also you can use `a:any-link`.

keygenqt: Fix

src/selection-tools.js
```diff
@@ -249,24 +249,50 @@ html, body {
...
     margin: 8px 0;
 }
+footer {
+    display: flex;
+    justify-content: space-between;
+    flex-direction: row;
+    width: 100%;
+}
+footer div {
+    white-space: nowrap;
```
johnfactotum: The text will not always fit. You might try setting the flex-basis to max-content or fit-content but still allow it to shrink and wrap. The same applies to the <a>. Also should probably add a gap so that the texts won't run together.

keygenqt: There are no such long names. I added a fix for correct output.

Screenshot from 2024-09-24 07-54-05

johnfactotum: It's very much possible that the line would not fit. The texts are localized and in other languages they could be much longer than they are in English. Also the font is set to `menu`, so the text could also become too long depending on the font set by the user.

keygenqt: No matter how long the text is, the layout will not float, and the user will find what he needs in the pop-up. I fixed it.

keygenqt: + fix long language name
+ fix big "by Google"

johnfactotum:
So in the end I decided to fix it myself with 02c5727abde5f40695f4ecd3d6b3dbb300d09b92. It uses GSettings instead of localStorage.

As for language selection, the main layout is the same as before:
image

But now it's possible to filter the languages:
image

PR #1414, opened 2024-10-20
Update Swedish translation

NickWick13:
I have updated the Swedish translation. See ZIP-file for translation file.


NickWick13:
[sv.zip](https://github.com/user-attachments/files/17450627/sv.zip)

PR #1446, opened 2024-12-01
Update Greek translation

gmou3:
A few translations get messed up from time to time.

There also exist some places where the UI remains in English for me.

In the library viewer:
- All Books
- Add Catalog...
- Library Menu (hover)

In the book viewer:
- Pin Sidebar (hover)
- Scrolled Mode (book's main menu)
- Reduce Animation (inside a book's font menu, under behavior)
- Copy Identifier (pop-up after choosing text)

### DIFF
--- a/po/el.po
+++ b/po/el.po
@@ -72,7 +72,6 @@ msgid "Custom"
 msgstr "Προσαρμοσμένο"
 
 #: src/annotations.js:592
-#, fuzzy
 msgid "JSON Files"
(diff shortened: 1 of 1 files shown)


johnfactotum:
They don't work because of the `#, fuzzy` line, which means that the translation was automatically generated by gettext. So one should remove that line from entries where the translation has been resolved by a translator.

gmou3:
Thanks, fixed!

PR #1468, opened 2025-01-03
Turn page using single touchscreen tap

andrzej-az:
This pr allows to turn the page by single touchscreen tap.  The reading area is divided vertically into two zones:
40% (left side): Tapping here navigates to the previous page.
60% (right side): Tapping here navigates to the next page.

tested on epub in "single page" and "scrolled" mode.
seems does not affect on other functionality
NOTE:  this change affects touchscreen only, the mouse/touchpad behavior is left "as is"

### DIFF
--- a/src/book-viewer.js
+++ b/src/book-viewer.js
@@ -260,7 +260,13 @@ GObject.registerClass({
                 dyLast = null
             },
         }))
-
+        let pressedX = 0;
+        this.#webView.add_controller(utils.connect(new Gtk.GestureClick(), {
+            'pressed': (gesture, n_press, x, y) => {
+                pressedX = x;
+            },
+            'released': (gesture, n_press, x, y) => { if (pressedX == x) setTimeout(() => {this.onTap(x,y)},100)}
+        }));
         const applyStyle = () => this.#applyStyle().catch(e => console.error(e))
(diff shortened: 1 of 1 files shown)


src/book-viewer.js
```diff
@@ -260,7 +260,13 @@ GObject.registerClass({
                 dyLast = null
             },
         }))
-
+        let pressedX = 0;
+        this.#webView.add_controller(utils.connect(new Gtk.GestureClick(), {
+            'pressed': (gesture, n_press, x, y) => {
+                pressedX = x;
+            },
+            'released': (gesture, n_press, x, y) => { if (pressedX == x) setTimeout(() => {this.onTap(x,y)},100)}
```
johnfactotum, line 268: Is there a reason why `setTimeout` is needed?

andrzej-az: As far as I remember, the delay is needed thanks to swipe behavior: without it 2 or more pages will be turned.

johnfactotum:
Probably tapping will need to be handled in WebView and not on the GTK side because it needs to handle cases where there are clickable elements on the page. Although currently it doesn't support scripting and form submission, there's still `<a>` and `<details>` elements that shouldn't trigger page turns when tapped.

PR #1547, opened 2025-07-04
Added Belarusian translation

zjyhutka:
Added translation of interface into Belarusian located in po/be.po, the line "be" is also added to the LINGUAS file.

Спадзяюся, што будзе карысна

### DIFF
--- a/po/LINGUAS
+++ b/po/LINGUAS
@@ -1,4 +1,5 @@
 ar
+be
 cs
 de
 el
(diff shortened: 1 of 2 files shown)


johnfactotum:
The build failed with
```
../data/../po/be.po:1206:2: syntax error
../data/../po/be.po:1207: end-of-line within string
```

It seems it's missing a `"`.

zjyhutka:
Hi! Sorry for broken pr, now i have fixed it by adding missing quote

PR #1557, opened 2025-07-27
app: Idle-inhibit when window is fullscreen

agx:
When reading in fullscreen one usually doesn't want the session to go idle (thus dim the screen and eventually blank it).

With this reading on mobile is a breeze, without it I'd have to remember to turn on/off "caffeine" mode.

Edit: I'm not much of JS person so please excuse if this doesn't look very JS-like.

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -54,6 +54,7 @@ const ApplicationWindow = GObject.registerClass({
     #library
     #bookViewer
     #stack = new Gtk.Stack()
+    #cookie
     constructor(params) {
         super(params)
         Object.assign(this, {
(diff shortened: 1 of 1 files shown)


agx:
eslint looks mostly happy now, I'd just need a hint what to do about the `_pspec` unused variable.

johnfactotum:
You can just leave it out. Or if you feel like it, make it a comment. So like
```
win => {
```
or
```
(win/*, _pspec*/) => {
```

But probably just leave it out. I don't think I've ever used the spec when connecting to signals.

agx:
Let's just drop it then, thanks.

PR #1566, opened 2025-08-20
Add "Books" to list of keywords in .desktop file

johnlorentzson:
Perhaps changing the "book" keyword that exists would be the better move. I don't mind either way, I just want to be able to type "books" into a system search and get books.

### DIFF
--- a/data/com.github.johnfactotum.Foliate.desktop.in
+++ b/data/com.github.johnfactotum.Foliate.desktop.in
@@ -10,7 +10,7 @@ Icon=com.github.johnfactotum.Foliate
 Terminal=false
 Type=Application
 # Translators: Search terms to find this application. Do NOT translate or localize the semicolons! The list MUST also end with a semicolon!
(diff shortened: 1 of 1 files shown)


johnfactotum:
Adding it is fine, I guess. But perhaps one would need to fix all the translations as well. That is, to replace all the current instances of `Ebook;Book;EPUB;Viewer;Reader;` in the `.po` files with `Ebook;Book;Books;EPUB;Viewer;Reader;`, to avoid disrupting the translations. Otherwise the keywords would stop being translated until they are each fixed by a translator, which might take a long time.

djinnalexio:
> to replace all the current instances of `Ebook;Book;EPUB;Viewer;Reader;`

Do you mean just doing a *Find/Replace* on all the `po` files and let translators update the translation later? If so, it's quickly done.

Also, I'm proposing this:

```diff
- Keywords=Ebook;Book;EPUB;Viewer;Reader;
+ Keywords=ebooks;books;epub;ereader;reader;viewer;library;
```

PR #1578, opened 2025-09-14
feat: Add intelligent "Meaning" tool with dictionary/translation auto-detection

nov3o:
## Summary
  • Replace separate Dictionary/Translate tools with unified "Meaning" tool
  • Auto-detect: single words → Dictionary, phrases → Translation
  • Fallback: Dictionary lookup fails → Translation mode available with language selection
  • Manual switching with Dictionary/Translate tabs

screencast-400px

### DIFF
--- a/src/selection-tools.js
+++ b/src/selection-tools.js
@@ -29,18 +29,22 @@ const getGoogleTranslateLanguages = utils.memoize(() => {
 })
 
 const tools = {
-    'dictionary': {
-        label: _('Dictionary'),
-        uri: 'foliate-selection-tool:///selection-tools/wiktionary.html',
-        run: (__, { text, lang }) => ({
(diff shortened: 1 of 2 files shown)


johnfactotum:
It's not clear to me what use case this would be advantageous for. Perhaps if you need to switch back and forth between different tools. It used to have a "tabbed" design like this, together with the Wikipedia tool, which is similar to the design used by Kindle. But I think the current design does have some advantage in its simplicity.

As for the implementation, for a design like this one it would probably be better to use the new `AdwToggleGroup` widget. Also for detecting multiple words one should probably use `Intl.Segmenter` or something like that. There are other problems as well, such as some labels seemingly not being translatable, and the HTML page is not added to the GResource list.

Issue #1310, opened 2024-05-01
List keyboard shortcuts


johnfactotum:
### Discussed in https://github.com/johnfactotum/foliate/discussions/1233

<div type='discussions-op-text'>

<sup>Originally posted by **gotmax23** January 19, 2024</sup>
There used to be a menu to list all the available keyboard shortcuts in Foliate v2. Is this still there in the GTK 4 rewrite? I cannot seem to find it. Are the keyboard shortcuts listed somewhere else?</div>

amano-kenji:
What are the existing keyboard shortcuts? Ctrl+F used to launch a text finder, but it doesn't do anything now.

johnfactotum:
> What are the existing keyboard shortcuts?

See https://github.com/johnfactotum/foliate/blob/479e0d5f3e2e7e5e6edffce3487eac86ce7fdfea/src/book-viewer.js#L677-L706

> Ctrl+F used to launch a text finder, but it doesn't do anything now.

See #1299.

PR #1607, opened 2025-11-13
feat: ✨ Added the keyboard shortcuts help page

djinnalexio:
Added a GTK4 keyboard shortcuts window in the book menu and tied it to `<ctrl>?`. Addresses issue #1310 and discussion #1233.

Note that while `GtkShortcutsWindow` and its children are marked as deprecated, the documentation also states that there isn't an alternative yet, and that an eventual new implementation will be for Gtk5.

### DIFF
--- a/com.github.johnfactotum.Foliate.json
+++ b/com.github.johnfactotum.Foliate.json
@@ -1,7 +1,7 @@
 {
     "app-id" : "com.github.johnfactotum.Foliate",
     "runtime" : "org.gnome.Sdk",
-    "runtime-version" : "48",
+    "runtime-version" : "49",
     "sdk" : "org.gnome.Sdk",
     "command" : "foliate",
     "finish-args" : [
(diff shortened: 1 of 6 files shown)


djinnalexio:
Addresses issue #1310 and discussion #1233

johnfactotum:
Libadwaita 1.8 includes the new ShortcutsDialog that can be used instead of GtkShortcutsWindow: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.ShortcutsDialog.html

djinnalexio:
> Libadwaita 1.8 includes the new ShortcutsDialog that can be used instead of GtkShortcutsWindow: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1-latest/class.ShortcutsDialog.html

Thanks. I upgraded to GNOME 49 to start working with it.

By the way, is there a reason `helpOverlay` seemed to be originally defined in `app.js` (accessed with `win.help-overlay`) rather than `book-viewer.js`? I think it fits better there since it's only relevant to the reader window (`viewer.help-overlay`).

johnfactotum:
> By the way, is there a reason `helpOverlay` seemed to be originally defined in `app.js`

It used to be that you would use the same GtkShortcutsWindow for different sections of an app. It seems this is no longer a thing in AdwShortcutsDialog. Which, frankly, makes no sense. But oh well.

PR #1610, opened 2025-11-22
Inhibit Idle when window is maximized

### LINKED ISSUE #1609: Inhibit system idling while viewing a document in maximized mode
**Is your feature request related to a problem? Please describe.**
Due to inactivity while reading, the display tries to turn off to save power.

**Describe the solution you'd like**
A sort of wake-lock acquired by the application, to prevent the display from turning off.

**Describe alternatives you've considered**
None

**Additional context**
System info: Fedora Workstation 42 with GNOME on Wayland

skr4n:
Currently, the application inhibits idle only for full-screen mode.
It would be helpful to have this when the window is maximized as well.

Closes #1609

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -83,12 +83,17 @@ const ApplicationWindow = GObject.registerClass({
 
 
         const app = Gio.Application.get_default()
-        const inhibitIdle = (win) => {
+        const toggleInhibitIdle = (win) => {
+            const canInhibit = this.is_active && (this.is_fullscreen() || this.is_maximized())
+            if (!canInhibit) {
+                uninhibitIdle()
+                return
+            }
             if (this.#inhibitCookie) return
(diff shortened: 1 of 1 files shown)


camelCaseNick:
I would highly suggest also checking for `this.is_active()` and `notify::is-active`, because otherwise when Foliate is maximised in the background, e.g. on a different workspace or behind a different window, it will still inhibit the session.

skr4n:
> I would highly suggest also checking for `this.is_active()` and `notify::is-active`, because otherwise when Foliate is maximised in the background, e.g. on a different workspace or behind a different window, it will still inhibit the session.

Fixed. Thanks for pointing it out.

LibMeow2025:
can this be merged please? @johnfactotum

Issue #1458, opened 2024-12-14
Multi-Select when importing Books


iNewmanity:
**Is your feature request related to a problem? Please describe.**
When I import my Books i only select one at a time. There is no possibility to select multiple ones in the file-browser

**Describe the solution you'd like**
Enable multi-select in the file browser.

IBBoard:
There's already a ticket for this with a command-line workaround - https://github.com/johnfactotum/foliate/issues/1196#issuecomment-2438425213

Jopp-gh:
Opening single PDFs in a eBook reader is simply impractical. 

Instead to add a "multiple file option", I'd prefer if Foliate could open a **folder chooser** to import recursively all PDFs / ebooks from a selected folder. 

To go a step further, Foliate could "watch a folder" and add new books automatically. 
Such features would be adequate for the quality standard Foliate strives for.
Watching a folder for changes is nothing special or incredible difficult to implement.

mrlem:
> There's already a ticket for this with a command-line workaround - [#1196 (comment)](https://github.com/johnfactotum/foliate/issues/1196#issuecomment-2438425213)

Maybe this issue can be closed as a duplicate then?

LolloDev5123:
I second this issue. It came like a shock to me that I couldn't import my folder of 200+ PDFs all at once.

Zain-Jaafar:
This issue is covered by pull request #1628, hopefully it gets approved

Issue #1196, opened 2023-12-26
Add multiple books to library at once


crimsonfall:
Currently the only way to add a book (.epub file) to your personal Foliate library is to open them, however you can only do so one-at-a-time (there's no way to select multiple books at once). This makes it frustrating if, for example, you already have a sizeable collection of ebooks you want to transfer over to Foliate, but would have to spend a considerable amount of time adding each book to your library. In previous versions of Foliate, you were able add multiple books to your library at once.

A solution to this issue would be of course to simply have a feature where you can add multiple ebooks (.epub files or other formats) to your library at once without opening them all. Perhaps a simple ‘Add’ or ‘+’ button would suffice, which upon pressing it, would open your file manager and allow you to select your ebook(s).

Edit: I forgot to mention earlier, but I think adding the ability to also remove multiple books from a library at once would also be nice.

(Congratulations on your app rewrite by the way, it looks and functions far better than before!)

jaybent:
Workaround:
`find . -type f -print0 | xargs -0 -I {} bash -c 'foliate "{}" & sleep 1' && sleep 10 && killall foliate`

Explanation:
This workaround allows opens multiple files with foliate and then close all instances of foliate at once, effectively bulk adding all files under current path (recursively).

1. `find . -print0` recursively finds all files from the current directory (the period), printing them separated by null characters.
2. `xargs -0 -I {} bash -c 'foliate "{}" & sleep 1'` uses `xargs` to pass each found item to the `bash` command:
   - `foliate "{}"` opens the file with `foliate` and runs it in the background.
   - `sleep 1` gives a brief pause to ensure the GUI window has time to load for each instance.
3. `&& sleep 10` waits for 10 seconds after all instances have been started, giving them time to fully load.
4. `killall foliate` terminates all instances of `foliate` at once.

Caveat:
The sleep timers (`sleep 1` and `sleep 10`) might need to be tweaked depending on your system and/or disk speed to ensure all windows have time to open properly, and for foliate to write all required files before being closed.

This assumes all files under current path are e-books, it is undefined what happens if any other files exists. Modify the `find` command if this is a problem, example: `find . -type f -iname *.pdf -print0` will only match files with names ending in `.pdf`, likely being PDFs.

Warning:
This command will open all files found, it could potentially consume a huge amount of RAM and crash the system if there is a lot of files or they are large.

IBBoard:
An alternative command that should use less RAM (but will take a little longer):

[...]

PR #1628, opened 2026-01-22
Import multiple books

attiqeurrehman:
This pull request introduces a new bulk import feature for e-books, allowing users to import multiple books at once from files or folders. It also includes several UI improvements and minor bug fixes to enhance stability and user experience.

**Bulk Import Feature:**

* Added a new `import-books` action to `ApplicationWindow`, enabling users to import multiple e-books via a dialog that supports both file and folder selection. This includes recursive folder scanning and feedback to the user via toast notifications. [[1]](diffhunk://#diff-c72a907ac323cd2f334ed0e2bd07d15ab62581c4753660c8a0d1c681b30be4b6L77-R77) [[2]](diffhunk://#diff-c72a907ac323cd2f334ed0e2bd07d15ab62581c4753660c8a0d1c681b30be4b6R176-R371)
* Registered a keyboard shortcut (`Ctrl+Shift+O`) for the new import action.
* Added an "Import Books…" menu item in the library UI, linked to the new action.

**Book Viewer and Library Stability Improvements:**

* Fixed a bug in `BookViewer` where the window title could be set when `root` was undefined, and improved cleanup logic in `vfunc_unroot`. [[1]](diffhunk://#diff-fee40a9c59d71e65bbf79ba70e28d160da0e7a6cee9e1e385594f8e0d131860cL745-R745) [[2]](diffhunk://#diff-fee40a9c59d71e65bbf79ba70e28d160da0e7a6cee9e1e385594f8e0d131860cR1007)
* In `Library`, ensured the sidebar header function is cleared during disposal to prevent callbacks after shutdown.

**UI Updates:**

* Replaced `AdwSpinner` with `GtkSpinner` and updated the contents stack switcher to use `AdwViewSwitcher` with improved policy settings in the book viewer UI. [[1]](diffhunk://#diff-6848d3db42ed6a7283fbaeb9d87ade7769fa53b8229b032b3e68b2c8df13882fL151-R151) [[2]](diffhunk://#diff-6848d3db42ed6a7283fbaeb9d87ade7769fa53b8229b032b3e68b2c8df13882fL285-R287)

### DIFF
--- a/src/app.js
+++ b/src/app.js
@@ -74,7 +74,7 @@ const ApplicationWindow = GObject.registerClass({
         this.connect('destroy', () => styleManager.disconnect(handler))
 
         utils.addMethods(this, {
-            actions: ['open', 'close', 'show-library', 'show-menu', 'new-window', 'open-copy'],
+            actions: ['open', 'close', 'show-library', 'show-menu', 'new-window', 'open-copy', 'import-books'],
             props: ['fullscreened'],
         })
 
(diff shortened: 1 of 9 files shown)


src/ui/book-viewer.ui
```diff
@@ -148,7 +148,7 @@
             <object class="GtkStackPage">
               <property name="name">loading</property>
               <property name="child">
-                <object class="AdwSpinner"/>
+                <object class="GtkSpinner"/>
```
IBBoard: Why is the ADW Spinner being swapped for a GTK spinner? What difference does it make? And why switch back to a GTK widget when the rest of the UI uses ADW widgets?

src/ui/book-viewer.ui
```diff
@@ -282,9 +282,9 @@
                 </object>
               </child>
               <child type="bottom">
-                <object class="AdwInlineViewSwitcher" id="contents-stack-switcher">
+                <object class="AdwViewSwitcher" id="contents-stack-switcher">
                   <property name="stack">contents-stack</property>
-                  <property name="display-mode">icons</property>
+                  <property name="policy">wide</property>
```
IBBoard: What has this got to do with importing books? And why is this an "improved policy setting"? The sidebar is normally narrow, so "icons" makes sense. Forcing wide mode on a narrow component doesn't seem helpful.

src/app.js
```diff
@@ -173,6 +173,202 @@ const ApplicationWindow = GObject.registerClass({
...
+        dialog.filters.append(ebooks)
+        dialog.default_filter = ebooks
+
+        dialog.open_multiple(this, null, (_, res) => {
+            try {
+                const files = dialog.open_multiple_finish(res)
+                let importedCount = 0
+                for (let i = 0; i < files.get_n_items(); i++) {
+                    const file = files.get_item(i)
+                    this.openFile(file)
```
IBBoard: It looks like this is literally going to open every single ebook that you select. Not just import them, but open and render them. That seems… inefficient! Can you not put them directly into the URI store?

src/app.js
```diff
@@ -173,6 +173,202 @@ const ApplicationWindow = GObject.registerClass({
...
+            name: _('E-Book Files'),
+            mime_types: [
+                'application/epub+zip',
+                'application/x-mobipocket-ebook',
+                'application/vnd.amazon.mobi8-ebook',
+                'application/x-mobi8-ebook',
+                'application/x-fictionbook+xml',
+                'application/x-zip-compressed-fb2',
+                'application/vnd.comicbook+zip',
+            ],
```
IBBoard: This seems to replicate the list in `open()`. It's easy for the lists to get out of sync. Now that it's required in two places then it would be better to have a single constant list and reference it in both places.

src/app.js
```diff
@@ -173,6 +173,202 @@ const ApplicationWindow = GObject.registerClass({
             }
         })
     }
+    #collectEbooksFromDirectory(directory, files = []) {
+        const ebookExtensions = ['.epub', '.mobi', '.azw', '.azw3', '.fb2', '.fb2.zip', '.cbz']
```
IBBoard: Rather than relying on file extensions here and mime types elsewhere, why not use [content_type_guess()](https://docs.gtk.org/gio/func.content_type_guess.html) to get a file type and use the existing list?

IBBoard (commented):
I'd like this feature to avoid using the bash script that I put in one of the existing "we need an import feature" tickets, so I thought I'd have a look and give some feedback.

attiqeurrehman:
Can you link the existing script so that and i can benefit from it. I will address your comments soon as well.

IBBoard:
The tickets are already linked. 

 #1458 has a comment from me about it being a dupe of https://github.com/johnfactotum/foliate/issues/1196

https://github.com/johnfactotum/foliate/issues/1196 had several different work-around scripts from different users, each of which is a variation on "open a file with Foliate so that it adds it to the database, then close it and open the next one"

src/ui/book-viewer.ui
attiqeurrehman: You're absolutely right - this was an unrelated change that shouldn't have been in this PR. I've reverted it back to AdwSpinner.

src/ui/book-viewer.ui
attiqeurrehman: Good catch! You're correct that forcing wide mode on a narrow sidebar doesn't make sense. I've reverted this to use AdwInlineViewSwitcher with display-mode="icons" as it was before.

src/app.js
attiqeurrehman: Excellent point! I've completely reworked the implementation. The books are now imported using the existing openFile() method which is what happens when users normally open books - this automatically adds them to the library. I've removed all the complex async processing, timeout handling, and GC management. The import is now much simpler and more efficient.

src/app.js
attiqeurrehman: Great suggestion! I've created a shared EBOOK_MIME_TYPES constant that's used by both open() and the import functions. This ensures the lists stay in sync.

src/app.js
attiqeurrehman: Excellent idea! I've implemented an `isEbookFile()` helper function that uses `Gio.content_type_guess()` to properly detect file types based on content rather than just extensions. This is much more robust and works with the shared `EBOOK_MIME_TYPES` constant.

src/app.js
IBBoard: Simplicity is good. But that doesn't actually address the point. You're still calling `openFile()` and (as far as I can see) that will still open and render _every single book_.

Also, it's weird that I commented on this line in `importMultipleFiles()` but you reworked `importFromFolder()`.

And there's now much more obvious duplication between the functions. Both are building a list of files and then invoking `openFile()` on them. Can you not build lists of files in the two functions (multi-file import and folder import) and pass that list to a common function?

src/app.js
```diff
@@ -254,37 +254,20 @@ const ApplicationWindow = GObject.registerClass({
...
-                    } catch (e) {
-                        console.error(`Failed to import ${file.get_path()}:`, e)
-                    }
-                }
                 
-                if (importedCount > 0) {
+                if (files.length > 0) {
+                    // Open each file - this adds them to the library
+                    files.forEach(file => this.openFile(file))
+
```
IBBoard: You've lost error handling now. What happens if one file in 500 fails to open?

IBBoard (commented):
It's different. I don't think that fixed anything, though. And error handling is now worse.

src/ui/book-viewer.ui
IBBoard: Why was it changed?

src/ui/book-viewer.ui
IBBoard: What was the thinking behind making it wide mode?

src/app.js
attiqeurrehman: yes on it.

src/ui/book-viewer.ui
attiqeurrehman: I am just learning about Gtk so there is knowledge gap.

attiqeurrehman: Apparently I was running `libadwaita-1 version 1.5.0`  that is causing the issues `AdwInlineViewSwitcher`, have to fix this first and then try with the original code

IBBoard: I know people have to learn new toolkits. I understand having issues with dependencies and not having your development environment set up correctly. I don't understand adding unrelated changes to a PR, removing them when someone points out that they're unrelated, unnecessary and a regression in behaviour, _and then re-adding them in the very next commit_.

If you can't get your dev environment set up correctly and need to have workarounds then you need to at least keep them out of your committed changes.

src/app.js
```diff
@@ -213,6 +234,70 @@ const ApplicationWindow = GObject.registerClass({
...
+            }
+            
+            // Open file in hidden viewer to extract metadata
+            hiddenViewer.open(file)
+            
+            // Continue with next file after a delay
+            GLib.timeout_add(GLib.PRIORITY_DEFAULT, 2000, () => {
+                importNext()
+                return GLib.SOURCE_REMOVE
+            })
```
IBBoard: You're still creating an ebook rendering widget and adding it to the stack widget. Yes, it has the side-effect of adding the book to the library, but it's still also doing all of the unnecessary UI stuff at the same time.

As a rule, if you're doing a bulk import and adding a timeout then you're probably doing it wrong and there's a better, more direct way of implementing it. Bulk imports should just be able to chain their operations one after another (or as async calls/promises where you wait for completion).

There's a `URIStore` in `library.js`. It has a `getURIStore` function that gets the library. Books are [added to the library](https://github.com/attiqeurrehman/foliate/blob/eb7ecb27acf9cee1c0ae68f1a5b4495f31090990/src/book-viewer.js#L794) when the book is loaded and emits a `book-ready` signal. That happens when Foliate calls the FoliateJS library to [make the book](https://github.com/attiqeurrehman/foliate/blob/eb7ecb27acf9cee1c0ae68f1a5b4495f31090990/src/reader/reader.js#L53-L57) and it finishes setting things up.

So rather than doing a whole rendering thing with GUI widgets then all you should need to do is:
* [`makeBook()`](https://github.com/attiqeurrehman/foliate/blob/eb7ecb27acf9cee1c0ae68f1a5b4495f31090990/src/reader/reader.js#L53) to read the book and its metadata 
* [Generate and assign an identifier](https://github.com/attiqeurrehman/foliate/blob/eb7ecb27acf9cee1c0ae68f1a5b4495f31090990/src/book-viewer.js#L775)
* [Save the book-specific metadata](https://github.com/attiqeurrehman/foliate/blob/eb7ecb27acf9cee1c0ae68f1a5b4495f31090990/src/book-viewer.js#L793) by creating a [JSONStorage object with the right values](https://github.com/attiqeurrehman/foliate/blob/eb7ecb27acf9cee1c0ae68f1a5b4495f31090990/src/data.js#L20-L25)
* Optionally [save the cover](https://github.com/attiqeurrehman/foliate/blob/gtk4/src/data.js#L83-L94) (if there is one), which [comes from the FoliateJS book object](https://github.com/attiqeurrehman/foliate/blob/gtk4/src/reader/reader.js#L487-L488)
* [Save the URI in the library](https://github.com/attiqeurrehman/foliate/blob/eb7ecb27acf9cee1c0ae68f1a5b4495f31090990/src/book-viewer.js#L794C13-L794C43), which is just a [thin wrapper around the URI store](https://github.com/attiqeurrehman/foliate/blob/gtk4/src/data.js#L95-L101)

(It's just a shame that Foliate has the code quite closely coupled between the UI and the storage and doesn't have a function that _just_ does the I/O parts)

johnfactotum:
There are many issues here. Mainly:
- Unrelated changes should be split into separate PRs
- Run `eslint` to make sure the style is consistent and avoid making unnecessary changes such as changing the indentation or adding semicolons.
- There's a lot here that's just redoing what's already handled in `src/utils.js`, etc.
- It should avoid reading the contents of the files directly. Rather, the trick is to use the WebView's file chooser, which is a bit of a hack, but makes things much more efficient.
- New files need to be added to the GResources so they'll get compiled.

For the main feature of importing books (only multiple files, not from directories), I've pushed d68c1d6f3bd0753e7866aa069c79e10ce12b6c77 myself. So I will close this for now. But feel free to take some of your other fixes and make new PRs.

